PR #557, opened 2024-06-13
Po 367

max-holland:
### Checklist

- [ ] commit messages are meaningful and follow good commit message guidelines
- [ ] tests have been updated / new tests has been added (if needed)


cypress/support/step_definitions/accountDetails/accountDetailsPage.ts
```diff
@@ -5,10 +5,13 @@ Then('should display business unit account I selected from Business unit and def
   cy.get('#accountDetailsBusinessUnitValue').should('have.text', 'Cambridgeshire');
```
CadeFaulkner, line 5: @rajanigandra this value is hard coded at the step-def level, Ideally either get this value from the previous screen or assign it in the feature file, only raising this as going forward when we are selecting the business unit its not always going to be 'Cambridgeshire'

CadeFaulkner (approved):

PR #2343, opened 2026-03-17
PO-2984 Adding href, removing redundant attributes and amending styles for links

ChrisGarratt122:
### Jira link

See [PO-2984](https://tools.hmcts.net/jira/browse/PO-2984)

### Change description

The cause of links rendering with incorrect styling (black colour as opposed to blue, and an IBeam cursor on hover when it should be a pointer) is caused when links are used without the `href=""` attribute. The presence of `href` on links is in the best interest of accessibility, but is only functionally required when navigating away from a page.
Ideally link elements would only be used when navigating away from a page and would always use `href` to achieve this, however there are many existing links across the application that are either used not to navigate away from the page or achieve their functionality via a `(click)` event.

In the interest of retaining existing functionality across the application but still prioritising accessibility and expected stying, this PR is a compromise that applies the following changes to links application wide:
- Ensuring the presence of `href`, using `href=""` where functionality is achieved via a click event.
- Ensuring `event.preventDefault()` is called in underlying methods to prevent unintended page navigation.
- Removing `tabindex="0"`. This is redundant as links with `href` are focusable by default.
- Removing redundant `(keydown.enter)`/`(keyup.enter)` handlers, links with `href` trigger click on enter by default.

### Testing done

Several affected links tested manually, ensured at least one instance of each different 'type' of link tested manually (e.g. list link, cancel link, delete link)
Tests introduced to check for presence of expected attributes on links and absence of redundant attributes.
Tests introduced to ensure original underlying methods are still being called and that `event.preventDefault()` is being called.

  * [ ] Yes
  * [x] No

### Checklist

- [x] README and other documentation has been updated / added (if needed)


iamfrankiemoran (changes requested):
Minor comments to address

iamfrankiemoran (approved):
LGTM

Arnabsubedi233 (approved):
LGTM

CadeFaulkner (approved):

PR #2680, opened 2026-06-08
PO-2305 Select business units for operational reports

TimDanielsCQI:
### Jira link

See [PO-2305 ](https://tools.hmcts.net/jira/browse/PO-2305)

### Change description

Most reports are against a single business unit. The 2 enforcement operational reports will be generated across multiple BUs, i.e. a single CSV file will be created and include data from all the BUs that the user selects. They may only select BUs in which they have the 'right' to that report in those business units.

- Navigation to the screen when 'Create Report' is selected from either of the reports' summary screens (by enforcement & by payments);
- Showing the relevant BUs, and allowing them to be selected;
- Warning dialog if x amount of BUs selected (see below)
- Continue to the screen that allows the user to enter parameters;
- Cancel the creation of the report.

### Backend Branch

This ticket was developed against the branch in this PR: https://github.com/hmcts/opal-fines-service/pull/1903 as it needs a system parameter and permission code not in master yet

### Testing done

- Unit tests are done
- Manual testing in the browser
- Linting and /Review codex skills done

  * [ ] Yes
  * [x] No

### Checklist


.gitignore
iamfrankiemoran, line 1: Undo the change here

src/app/flows/fines/services/opal-fines-service/interfaces/opal-fines-report.interface.ts
```diff
@@ -0,0 +1,18 @@
+import { IOpalFinesReportParameters } from './opal-fines-report-parameters.interface';
+
+export type TOpalFinesSupportedFileType = 'CSV' | 'PDF' | 'XML';
```
iamfrankiemoran: Can we move this to a types folder please in its own file

src/app/flows/fines/fines-sa/fines-sa-search/fines-sa-search-filter-business-unit/validators/fines-sa-search-filter-business-unit-select-bu.validator.ts
```diff
@@ -1,28 +1,4 @@
...
-  (group: AbstractControl): ValidationErrors | null => {
-    if (!(group instanceof FormGroup)) return null;
-    const record = group.get(recordControlName);
-    if (!record) return null;
-    return record.errors?.['required'] ? { [recordControlName]: { required: true } } : null;
-  };
+export {
+  atLeastOneBusinessUnitSelectedRecordValidator,
+  businessUnitSelectionRootMirrorValidator,
+} from '@app/flows/fines/validators/business-unit-selection.validator';
```
iamfrankiemoran: Why? We'd just delete this file and import them where we need it

src/app/flows/fines/fines-reports/utils/get-fines-reports-selected-business-unit-ids-from-navigation-state.util.ts
iamfrankiemoran, line 1: We never use browser history state at all

src/app/flows/fines/fines-reports/routing/constants/fines-reports-routing-paths.constant.ts
```diff
@@ -4,5 +4,8 @@ export const FINES_REPORTS_ROUTING_PATHS: IFinesReportsRoutingPaths = {
   root: 'reports',
   children: {
     summaryList: 'summary-list',
+    selectBusinessUnits: 'create/select-business-unit',
+    businessUnitWarning: 'create/business-unit-warning',
+    parameters: 'create/report-parameters',
```
iamfrankiemoran: Should create not have its own routing file? Where the root is 'create' with these?

src/app/flows/fines/fines-reports/interfaces/fines-reports-definition.interface.ts
```diff
@@ -0,0 +1,9 @@
+export interface IFinesReportsDefinition {
+  id: string;
+  heading: string;
+  title: string;
+  permissionIds: readonly number[];
+  operationalLinkId?: string;
+  highlightLinkId?: string;
+  highlightLinkText?: string;
+}
```
iamfrankiemoran: `string | null` instead of optional please

src/app/flows/fines/fines-reports/interfaces/fines-reports-business-unit-navigation-state.interface.ts
```diff
@@ -0,0 +1,3 @@
+export interface IFinesReportsBusinessUnitNavigationState {
+  selectedBusinessUnitIds?: number[];
```
iamfrankiemoran: Again, not sure if this should be `number[] | null` rather than optional

src/app/flows/fines/fines-reports/fines-reports-select-business-units/fines-reports-select-business-units.component.ts
```diff
@@ -0,0 +1,137 @@
...
+
+@Component({
+  selector: 'app-fines-reports-select-business-units',
+  imports: [GovukHeadingWithCaptionComponent, FinesReportsSelectBusinessUnitsFormComponent],
+  templateUrl: './fines-reports-select-business-units.component.html',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class FinesReportsSelectBusinessUnitsComponent extends AbstractFormParentBaseComponent implements OnInit {
+  private readonly route = inject(ActivatedRoute);
+  private readonly location = inject(Location);
```
iamfrankiemoran: Why?

src/app/flows/fines/fines-reports/fines-reports-select-business-units/fines-reports-select-business-units.component.ts
```diff
@@ -0,0 +1,137 @@
...
+   */
+  public ngOnInit(): void {
+    const resolverData = this.route.snapshot.data['businessUnits'] as IOpalFinesBusinessUnitRefData | undefined;
+
+    this.businessUnits = [...(resolverData?.refData ?? [])].sort((left, right) =>
+      left.business_unit_name.localeCompare(right.business_unit_name),
+    );
+    this.selectedBusinessUnitIds = getFinesReportsSelectedBusinessUnitIdsFromNavigationState(
+      this.routerService,
+      this.location,
```
iamfrankiemoran: Not sure on this

src/app/flows/fines/fines-reports/interfaces/fines-reports-definition.interface.ts
iamfrankiemoran: Could we not have extended the interface from the common ui lib?

src/app/flows/fines/constants/reports-primary-navigation-permissions.constant.ts
```diff
@@ -1,6 +1,8 @@
 import { FINES_PERMISSIONS } from '@app/constants/fines-permissions.constant';
 
-export const REPORTS_PERMISSIONS = [
```
louisbriggs, line 3: Can these be split into separate files, one for each const.

src/app/flows/fines/fines-reports/constants/fines-reports-definitions.constant.ts
```diff
@@ -0,0 +1,33 @@
+import { IFinesReportsDefinition } from '../interfaces/fines-reports-definition.interface';
+import { FINES_REPORTS_SUMMARY_LIST_ROUTING_PATHS } from '../fines-reports-summary-list/routing/constants/fines-reports-summary-list-routing-paths.constant';
+import { FINES_REPORTS_SUMMARY_LIST_ROUTING_TITLES } from '../fines-reports-summary-list/routing/constants/fines-reports-summary-list-routing-titles.constant';
+import { OPERATIONAL_REPORT_ROUTE_PERMISSIONS } from '../../constants/reports-permissions.constant';
+
+export const FINES_REPORTS_DEFINITIONS: IFinesReportsDefinition[] = [
```
louisbriggs: Split into individual files.

src/app/flows/fines/fines-reports/constants/fines-reports-highlight-report-links.constant.ts
```diff
@@ -1,22 +1,22 @@
...
 
-export const FINES_REPORTS_HIGHLIGHT_REPORT_LINKS: IDashboardPageConfigurationLink[] = [
-  {
-    id: 'user-reports',
-    text: 'View all your reports',
-    routerLink: [
-      '/',
-      FINES_ROUTING_PATHS.root,
-      FINES_REPORTS_ROUTING_PATHS.root,
-      FINES_REPORTS_SUMMARY_LIST_ROUTING_PATHS.children.yourReports,
```
louisbriggs, line 14: Can this usage be explained in comments.

src/app/flows/fines/fines-reports/fines-reports-business-unit-warning/fines-reports-business-unit-warning.component.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  public selectedBusinessUnitIds: number[] = [];
+
+  /**
+   * Returns the warning heading using the current selected business unit count.
+   *
+   * @returns The warning heading shown on the business unit warning page.
+   */
+  public get warningHeading(): string {
+    return `You have selected ${this.selectedBusinessUnitIds.length} business units`;
```
louisbriggs: Can this a const or similar to fines-reports-operational-report-links.constant.ts‎ in terms of dynamically injecting the BU length.

src/app/flows/fines/fines-reports/fines-reports-parameters/fines-reports-parameters.component.ts
```diff
@@ -0,0 +1,69 @@
...
+    this.router.navigate([`../../${FINES_REPORTS_ROUTING_PATHS.children.selectBusinessUnits}`], {
+      relativeTo: this.activatedRoute,
+    });
+  }
+
+  /**
+   * Initialises the placeholder parameters screen using the selected business units passed through navigation state.
+   */
+  public ngOnInit(): void {
+    const selectedBusinessUnitIds = getFinesReportsSelectedBusinessUnitIdsFromNavigationState(
```
louisbriggs: Do we need to use location here?

src/app/flows/fines/fines-reports/fines-reports-select-business-units/fines-reports-select-business-units-form/fines-reports-select-business-units-form.component.ts
```diff
@@ -0,0 +1,280 @@
...
+
+  /**
+   * Form control name for the master select-all business unit checkbox.
+   */
+  private readonly ALL_BUSINESS_UNITS_CTRL = 'fines_reports_select_business_unit_ids_select_all';
+
+  /**
+   * Signal storing the latest checked state for each business unit id.
+   */
+  private readonly businessUnitSelections = signal<Record<number, boolean>>({});
```
louisbriggs: Are signals required here?

src/app/flows/fines/fines-reports/fines-reports-select-business-units/fines-reports-select-business-units-form/fines-reports-select-business-units-form.component.ts
```diff
@@ -0,0 +1,280 @@
...
+
+  protected override fieldErrors: IAbstractFormBaseFieldErrors = {
```
louisbriggs: Field errors can be moved to it's own const.

src/app/flows/fines/fines-reports/fines-reports-select-business-units/fines-reports-select-business-units-form/fines-reports-select-business-units-form.component.ts
```diff
@@ -0,0 +1,280 @@
...
+      },
+    },
+  };
+
+  protected override formSubmit = new EventEmitter<IAbstractFormBaseForm<IFinesReportsSelectBusinessUnitsFormState>>();
+
+  /**
+   * Emits whether the selection form currently contains unsaved changes.
+   */
+  @Output() protected override unsavedChanges = new EventEmitter<boolean>();
```
louisbriggs: Does the abstract class unsavedChanges handle this already?

src/app/flows/fines/fines-reports/fines-reports-select-business-units/fines-reports-select-business-units-form/fines-reports-select-business-units-form.component.ts
```diff
@@ -0,0 +1,280 @@
...
+
+  /**
+   * Emits when the user selects the cancel link.
+   */
+  @Output() public cancelSelection = new EventEmitter<void>();
+
+  /**
+   * Id used by the master business unit checkbox.
+   */
+  public readonly allBusinessUnitsInputId = this.BUSINESS_UNITS_CTRL;
```
louisbriggs, line 88: does this need to be pre-defined or can just directly assign here.

src/app/flows/fines/fines-reports/fines-reports-select-business-units/fines-reports-select-business-units-form/fines-reports-select-business-units-form.component.ts
```diff
@@ -0,0 +1,280 @@
...
+  public readonly isAllSelected = computed(
+    () => this.businessUnits.length > 0 && this.selectedCount() === this.businessUnits.length,
+  );
+
+  /**
+   * Creates checkbox form controls keyed by business unit id.
+   *
+   * @returns A form record containing one unchecked control for each business unit.
+   */
+  private createBusinessUnitCheckboxControls(): FormRecord<FormControl<boolean>> {
```
louisbriggs: Check this method for consistency across app.

src/app/flows/fines/fines-reports/fines-reports-select-business-units/fines-reports-select-business-units-form/fines-reports-select-business-units-form.component.ts
```diff
@@ -0,0 +1,280 @@
...
+      validators: atLeastOneBusinessUnitSelectedRecordValidator,
+    });
+  }
+
+  /**
+   * Builds the row view model used by the template.
+   *
+   * @param record - The business unit form record that owns each row control.
+   */
+  private buildBusinessUnitRows(record: FormRecord<FormControl<boolean>>): void {
```
louisbriggs: Re assess the flow of creating the form controls.

src/app/flows/fines/fines-reports/fines-reports-select-business-units/fines-reports-select-business-units-form/fines-reports-select-business-units-form.component.ts
```diff
@@ -0,0 +1,280 @@
+import {
```
louisbriggs: Compare this component to a similar implementation in fines-sa-search-filter-business-unit-form.component

src/app/flows/fines/fines-reports/fines-reports-select-business-units/fines-reports-select-business-units.component.ts
```diff
@@ -0,0 +1,137 @@
+import { Location } from '@angular/common';
```
louisbriggs: Compare this to similar implementation here:- fines-sa-search-filter-business-unit.component

src/app/flows/fines/fines-reports/fines-reports-select-business-units/fines-reports-select-business-units.component.ts
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  /**
+   * Populates the current business unit list from the route resolver, sorts it alphabetically,
+   * and restores any in-progress selections.
+   */
+  public ngOnInit(): void {
+    const resolverData = this.route.snapshot.data['businessUnits'] as IOpalFinesBusinessUnitRefData | undefined;
+
+    this.businessUnits = [...(resolverData?.refData ?? [])].sort((left, right) =>
```
louisbriggs: Better to section out logic into their own methods.

src/app/flows/fines/fines-reports/fines-reports-summary-list/constants/fines-reports-summary-list-report-configuration.constant.ts
```diff
@@ -1,25 +1,10 @@
...
-    permissionIds: [FINES_PERMISSIONS['operational-report-by-enforcement']],
-  },
-  {
-    id: FINES_REPORTS_SUMMARY_LIST_ROUTING_PATHS.children.operationalReportsByPayments,
-    heading: 'Operational reports (by payments)',
-    title: FINES_REPORTS_SUMMARY_LIST_ROUTING_TITLES.children.operationalReportsByPayments,
-    permissionIds: [FINES_PERMISSIONS['operational-report-by-payments']],
-  },
-];
+export const FINES_REPORT_SUMMARY_LIST_REPORT_CONFIGURATION: IFinesReportSummaryListReportConfiguration[] =
```
louisbriggs: Does this need to be changed? Overly complex or required?

src/app/flows/fines/fines-reports/fines-reports-summary-list/interfaces/fines-report-summary-list-report-configuration.interface.ts
```diff
@@ -2,5 +2,5 @@ export interface IFinesReportSummaryListReportConfiguration {
   id: string;
   heading: string;
   title: string;
-  permissionIds: number[];
+  permissionIds: readonly number[];
```
louisbriggs: Is this needed?

src/app/flows/fines/fines-reports/fines-reports-summary-list/fines-reports-summary-list.component.ts
```diff
@@ -1,29 +1,71 @@
 import { ChangeDetectionStrategy, Component, inject } from '@angular/core';
```
louisbriggs, line 1: This is also being worked on in another PO-2307 so is likely to change.

src/app/flows/fines/fines-reports/interfaces/fines-reports-business-unit-navigation-state.interface.ts
louisbriggs: Does this need to be optional.

src/app/flows/fines/fines-reports/interfaces/fines-reports-definition.interface.ts
```diff
@@ -0,0 +1,9 @@
+export interface IFinesReportsDefinition {
+  id: string;
+  heading: string;
+  title: string;
+  permissionIds: readonly number[];
+  operationalLinkId?: string;
```
louisbriggs: Need to be optional?

src/app/flows/fines/fines-reports/routing/constants/fines-reports-routing-paths.constant.ts
```diff
@@ -4,5 +4,8 @@ export const FINES_REPORTS_ROUTING_PATHS: IFinesReportsRoutingPaths = {
   root: 'reports',
   children: {
     summaryList: 'summary-list',
+    selectBusinessUnits: 'create/select-business-unit',
```
louisbriggs: Create a constant just for the 'create' children instead of joining the paths here.

src/app/flows/fines/fines-reports/utils/get-fines-reports-selected-business-unit-ids-from-navigation-state.util.ts
```diff
@@ -0,0 +1,23 @@
...
+import { IFinesReportsBusinessUnitNavigationState } from '../interfaces/fines-reports-business-unit-navigation-state.interface';
+
+/**
+ * Reads selected business unit ids from the current router navigation or browser history state.
+ *
+ * @param router - Router used to access the current navigation extras state.
+ * @param location - Location used to access browser history state when navigation extras are unavailable.
+ * @returns Selected business unit ids, or an empty array when none were supplied.
+ */
+export const getFinesReportsSelectedBusinessUnitIdsFromNavigationState = (
```
louisbriggs: check where this is used and if it's required.

src/app/flows/fines/fines-sa/fines-sa-search/fines-sa-search-filter-business-unit/validators/fines-sa-search-filter-business-unit-select-bu.validator.ts
```diff
@@ -1,28 +1,4 @@
-import { AbstractControl, FormGroup, ValidationErrors, ValidatorFn } from '@angular/forms';
-
```
louisbriggs, line 2: This is a shim?

src/app/flows/fines/services/opal-fines-service/interfaces/opal-fines-cache.interface.ts
```diff
@@ -44,4 +45,5 @@ export interface IOpalFinesCache {
   minorCreditorAccountCreditorCache$: Observable<IOpalFinesAccountMinorCreditorCreditor> | null;
   minorCreditorAccountHistoryAndNotesCache$: Observable<IOpalFinesAccountMinorCreditorDetailsHistoryAndNotesTabRefData> | null;
   enforcersCache$: Observable<IOpalFinesEnforcersRefData> | null;
+  reportsCache$: { [key: string]: Observable<IOpalFinesReport> };
```
louisbriggs, line 50: check this is being cleared.

src/app/flows/fines/services/opal-fines-service/interfaces/opal-fines-report-parameters.interface.ts
```diff
@@ -0,0 +1,3 @@
+export interface IOpalFinesReportParameters extends Record<string, unknown> {
```
louisbriggs, line 1: This may not be required.

src/app/flows/fines/services/opal-fines-service/interfaces/opal-fines-report.interface.ts
louisbriggs: Split into separate files.

src/app/flows/fines/services/opal-fines-service/opal-fines.service.ts
```diff
@@ -251,6 +252,22 @@ export class OpalFines {
     return this.cache.businessUnitsCache$;
   }
 
+  /**
+   * Retrieves a report definition by report id and caches it for later route transitions.
+   *
+   * @param reportId - The report id to fetch from the reports API.
+   * @returns An observable containing the report definition.
+   */
+  public getReport(reportId: string): Observable<IOpalFinesReport> {
```
louisbriggs: check if this is or can be cleared.

louisbriggs (changes requested):
Suggested changes and improvements.

src/app/flows/fines/constants/reports-primary-navigation-permissions.constant.ts
TimDanielsCQI: In their own files now

src/app/flows/fines/fines-reports/constants/fines-reports-definitions.constant.ts
TimDanielsCQI: In their own files now

src/app/flows/fines/fines-reports/constants/fines-reports-highlight-report-links.constant.ts
TimDanielsCQI: I have reverted this file to how it was before I changed it. It was overly complicated, trying to dynamically build the "View all your reports" link from the FINES_REPORTS_DEFINITIONS, when there was no need really.

src/app/flows/fines/fines-reports/fines-reports-select-business-units/fines-reports-select-business-units-form/fines-reports-select-business-units-form.component.ts
TimDanielsCQI: Now in own constant file

src/app/flows/fines/fines-reports/fines-reports-select-business-units/fines-reports-select-business-units-form/fines-reports-select-business-units-form.component.ts
TimDanielsCQI: Yes it does so have removed this

src/app/flows/fines/fines-reports/fines-reports-select-business-units/fines-reports-select-business-units-form/fines-reports-select-business-units-form.component.ts
TimDanielsCQI: I've left BUSINESS_UNITS_CTRL in place because it’s still used in the form setup and validator wiring.

src/app/flows/fines/fines-reports/fines-reports-select-business-units/fines-reports-select-business-units-form/fines-reports-select-business-units-form.component.ts
TimDanielsCQI: The way I have done it seems broadly similar to **fines-sa-search-filter-business-unit-form.component** so I have left it as is.

src/app/flows/fines/fines-reports/fines-reports-select-business-units/fines-reports-select-business-units-form/fines-reports-select-business-units-form.component.ts
TimDanielsCQI: I have simplified this

src/app/flows/fines/fines-reports/fines-reports-select-business-units/fines-reports-select-business-units.component.ts
TimDanielsCQI: In its own helper method now

src/app/flows/fines/fines-reports/fines-reports-summary-list/constants/fines-reports-summary-list-report-configuration.constant.ts
TimDanielsCQI: Have put this file back to how it was, and simplified the code

src/app/flows/fines/fines-reports/fines-reports-summary-list/interfaces/fines-report-summary-list-report-configuration.interface.ts
TimDanielsCQI: This has gone now - I clone the constant OPERATIONAL_REPORT_ROUTE_PERMISSIONS in fines-reports-summary-list-report-configuration.constant.ts instead now

src/app/flows/fines/fines-reports/fines-reports-summary-list/fines-reports-summary-list.component.ts
TimDanielsCQI: I've kept this as minimal code to make this ticket work but may need to liaise with the developer of PO-2307 in order to avoid merge conflicts. I have got codex to do a test merge, and report on what to keep or override. The merge conflicts will need manually merging, but don't look too bad to unpick.

src/app/flows/fines/fines-reports/interfaces/fines-reports-definition.interface.ts
TimDanielsCQI: Making it string | null as per Frankie's comment

src/app/flows/fines/fines-reports/routing/constants/fines-reports-routing-paths.constant.ts
TimDanielsCQI: The route structure is now split so create is the parent and select-business-unit and business-unit-warning live underneath it

src/app/flows/fines/fines-sa/fines-sa-search/fines-sa-search-filter-business-unit/validators/fines-sa-search-filter-business-unit-select-bu.validator.ts
TimDanielsCQI: On reviewing this, it was a pointless wrapper and has been simplified now

src/app/flows/fines/services/opal-fines-service/interfaces/opal-fines-cache.interface.ts
TimDanielsCQI: It wasn't being cleared but it is now

src/app/flows/fines/services/opal-fines-service/opal-fines.service.ts
TimDanielsCQI: It is cleared now (see above comment)

src/app/flows/fines/services/opal-fines-service/interfaces/opal-fines-report-parameters.interface.ts
TimDanielsCQI: This is needed as business_unit_warning_threshold comes from the GET /opal-fines-service/reports/{id} response, inside report_parameters

TimDanielsCQI:
I have made all the required changes. 

I also added a **fines-reports.store.ts** to store:

s**electedReportTypeId
selectedBusinessUnitIds**

That means the selected business units are now held in signals state, not browser history/router state.

I removed the **fines-reports-parameters** package because it was only a placeholder used to prove the Continue action carried the selected business units forward. The selected ids are now stored in FinesReportsStore, ready for the future parameters screen to consume. The code logs them to the console.

src/app/flows/fines/services/opal-fines-service/types/opal-fines-supported-file-type.type.ts
iamfrankiemoran, line 1: Why is this in the interfaces folder? Should it not be in the types folder as it is a type?

src/app/flows/fines/routing/fines.routes.ts
```diff
@@ -152,6 +153,7 @@ export const finesRouting: Routes = [
         path: FINES_ROUTING_PATHS.children.reports.root,
         loadComponent: () => import('../fines-reports/fines-reports.component').then((c) => c.FinesReportsComponent),
         children: reportingRouting,
+        providers: [FinesReportsStore],
```
iamfrankiemoran, line 156: Why does the store need provided here?

src/app/flows/fines/fines-reports/stores/fines-reports.store.ts
```diff
@@ -0,0 +1,33 @@
+import { computed } from '@angular/core';
+import { patchState, signalStore, withComputed, withMethods, withState } from '@ngrx/signals';
+import { IFinesReportsState } from './interfaces/fines-reports-state.interface';
+
+export const FinesReportsStore = signalStore(
+  withState<IFinesReportsState>(() => ({
+    selectedReportTypeId: null,
+    selectedBusinessUnitIds: [],
+  })),
```
iamfrankiemoran, line 14: Might be better to put this into a constant as it is also repeated in the `clearSelectedBusinessUnitIds()`. It could also be used in the `setSelectedBusinessUnitIds()` by spreading in the constant

src/app/flows/fines/fines-reports/stores/interfaces/fines-reports-state.interface.ts
iamfrankiemoran, line 1: If we create a empty constant for this something like: 
```
export constant FINES_REPORT_STATE: IFinesReportsState {
  selectedReportTypeId: null,
  selectedBusinessUnitIds: [],
}
```

src/app/flows/fines/fines-reports/fines-reports-select-business-units/fines-reports-select-business-units.component.html
```diff
@@ -0,0 +1,15 @@
+<div class="govuk-grid-column-two-thirds">
+  <opal-lib-govuk-heading-with-caption
+    captionText="Create report"
+    [headingText]="reportHeading"
+  ></opal-lib-govuk-heading-with-caption>
+  <h2 class="govuk-heading-m">{{ pageHeading }}</h2>
+
```
iamfrankiemoran: I might be wrong, but I reckon this should be in the form component 🤔

src/app/flows/fines/fines-reports/fines-reports-select-business-units/fines-reports-select-business-units-form/fines-reports-select-business-units-form.component.html
```diff
@@ -0,0 +1,59 @@
...
+                [control]="row.control"
+              ></div>
+            </td>
+          </tr>
+        }
+      } @else {
+        <tr opal-lib-govuk-table-body-row>
+          <td id="business-units-empty" opal-lib-govuk-table-body-row-data>Business units will appear here.</td>
+        </tr>
+      }
```
iamfrankiemoran: Whats this for?

src/app/flows/fines/fines-reports/constants/fines-reports-operational-report-links.constant.ts
iamfrankiemoran, line 1: Trying to understand why this file was changed 🤔

src/app/flows/fines/fines-reports/constants/fines-reports-definitions.constant.ts
iamfrankiemoran, line 1: Trying to understand why this isn't just in `src/app/flows/fines/fines-reports/constants/fines-reports-operational-report-links.constant.ts` 🤔

src/app/flows/fines/fines-reports/constants/find-fines-reports-definition.constant.ts
iamfrankiemoran, line 1: What is this for?

src/app/flows/fines/constants/reports-primary-navigation-permissions.constant.ts
iamfrankiemoran, line 1: Why was this changed?

src/app/flows/fines/constants/operational-report-route-permissions.constant.ts
iamfrankiemoran, line 1: Why? Wherever you are using `OPERATIONAL_REPORT_ROUTE_PERMISSIONS` you could just use `FINES_PERMISSIONS['search-and-view-accounts']`

src/app/flows/fines/fines-reports/fines-reports-select-business-units/fines-reports-select-business-units-form/fines-reports-select-business-units-form.component.ts
```diff
@@ -0,0 +1,205 @@
...
+  }
+
+  /**
+   * Creates the reactive form and wires checkbox changes into component state.
+   */
+  private initialiseBusinessUnitForm(): void {
+    const record =
+      this.businessUnits.length > 1
+        ? this.createBusinessUnitCheckboxControls()
+        : new FormRecord<FormControl<boolean>>({});
```
iamfrankiemoran: [P2] Keep an empty business-unit result invalid

This branch treats both one and zero business units as a valid form. If the permission-filtered resolver returns an empty `refData` array, the empty `FormRecord` has no required validator, so Continue submits successfully and stores `[]`. Only bypass selection when `length === 1`; keep the form invalid or disable submission when there are zero units.

iamfrankiemoran (changes requested):

src/app/flows/fines/fines-sa/fines-sa-search/fines-sa-search-filter-business-unit/validators/fines-sa-search-filter-business-unit-select-bu.validator.ts
TimDanielsCQI: This was dealt with as part of Louis's comments

src/app/flows/fines/services/opal-fines-service/interfaces/opal-fines-report.interface.ts
TimDanielsCQI: Moved to types folder. Commit: [32906b94de73936ac5385573285436dc741ff85a](https://github.com/hmcts/opal-frontend/pull/2680/changes/32906b94de73936ac5385573285436dc741ff85a)

src/app/flows/fines/fines-reports/utils/get-fines-reports-selected-business-unit-ids-from-navigation-state.util.ts
TimDanielsCQI: I have already removed any use of browser history state responding to Louis's comments. I have created a fines-reports.store.ts to store **selectedReportTypeId** and **selectedBusinessUnitIds** because those values are flow state:

- **selectedReportTypeId** is because the create-report journey needs to know which report the user is creating after they leave the summary-list screen. It acts as the route context for the flow, and guards use it to stop people deep-linking into create steps without coming from View Reports.

- **selectedBusinessUnitIds** need to be stored because once the user leaves the select-business-units page, we still need those selections for the next step in the flow

Commit: https://github.com/hmcts/opal-frontend/pull/2680/changes/ea395705bad708ee0b01ca5cdcc669adad73235f

.gitignore
TimDanielsCQI: done

src/app/flows/fines/fines-reports/routing/constants/fines-reports-routing-paths.constant.ts
TimDanielsCQI: This has already been changed in response to a Louis comment.

Commit: https://github.com/hmcts/opal-frontend/pull/2680/changes/3873fd36344562d6095ea9df377af46785b9112b

src/app/flows/fines/fines-reports/interfaces/fines-reports-definition.interface.ts
TimDanielsCQI: Have simplified this area and this file is now gone

Commit: https://github.com/hmcts/opal-frontend/pull/2680/changes/656e2b943cbe6ba677df6def25038a773f2129e6

src/app/flows/fines/fines-reports/constants/fines-reports-definitions.constant.ts
TimDanielsCQI: I have simplified this and this file is now gone.

Commit: https://github.com/hmcts/opal-frontend/pull/2680/changes/656e2b943cbe6ba677df6def25038a773f2129e6

src/app/flows/fines/constants/operational-report-route-permissions.constant.ts
TimDanielsCQI: Simplification made.

Commit: [refactor: inline operational report permission](https://github.com/hmcts/opal-frontend/pull/2680/changes/2976d810b5c08ddf08aaf95f4331aa13c3752268)

src/app/flows/fines/fines-reports/interfaces/fines-reports-definition.interface.ts
TimDanielsCQI: Simplified code - this file gone now

src/app/flows/fines/fines-reports/fines-reports-business-unit-warning/fines-reports-business-unit-warning.component.ts
TimDanielsCQI: This now a constant.

Commit: [refactor: centralise business unit warning content](https://github.com/hmcts/opal-frontend/pull/2680/changes/4c29ebdcad656287c62fd0dad4476ca3c8f1ff2c)

src/app/flows/fines/fines-reports/fines-reports-parameters/fines-reports-parameters.component.ts
TimDanielsCQI: This component has gone - it was just to prove that the correct params were carried forward on the flow, but as you said, it creates tech debt if it is not in the scope of this Jira ticket

src/app/flows/fines/services/opal-fines-service/interfaces/opal-fines-report.interface.ts
TimDanielsCQI: Done

src/app/flows/fines/fines-reports/interfaces/fines-reports-business-unit-navigation-state.interface.ts
TimDanielsCQI: This interface was removed as part of the reports-state refactor and has no remaining usages.

src/app/flows/fines/fines-reports/fines-reports-select-business-units/fines-reports-select-business-units.component.ts
TimDanielsCQI: This has been dealt with during a refactor to not use browser history, but use a store instead

src/app/flows/fines/fines-reports/fines-reports-select-business-units/fines-reports-select-business-units.component.ts
TimDanielsCQI: This has been dealt with during a refactor to not use browser history, but use a store instead

src/app/flows/fines/services/opal-fines-service/types/opal-fines-supported-file-type.type.ts
TimDanielsCQI: In the types folder now

src/app/flows/fines/routing/fines.routes.ts
TimDanielsCQI: The store is route-provided so one instance is shared by the reports summary, create journey components and finesReportsStateGuard. It holds temporary selected-business-unit state needed when moving between those routes. Scoping it to /fines/reports means the state is cleared when the user leaves the Reports feature, rather than becoming application-wide state.

src/app/flows/fines/fines-reports/fines-reports-select-business-units/fines-reports-select-business-units-form/fines-reports-select-business-units-form.component.ts
TimDanielsCQI: This has gone now

src/app/flows/fines/fines-reports/fines-reports-select-business-units/fines-reports-select-business-units-form/fines-reports-select-business-units-form.component.ts
TimDanielsCQI: Have compared and it is broadly similar

src/app/flows/fines/fines-reports/fines-reports-select-business-units/fines-reports-select-business-units.component.ts
TimDanielsCQI: Have compared and it is broadly similar

src/app/flows/fines/fines-reports/interfaces/fines-reports-business-unit-navigation-state.interface.ts
TimDanielsCQI: This file has gone now during a simplification refactor

src/app/flows/fines/fines-reports/utils/get-fines-reports-selected-business-unit-ids-from-navigation-state.util.ts
TimDanielsCQI: This was dealt with during the refactor to get rid of using browser history

src/app/flows/fines/fines-reports/stores/fines-reports.store.ts
TimDanielsCQI: Done with commit:https://github.com/hmcts/opal-frontend/pull/2680/changes/73d807c079d2556fd4447cd486ce2639fed2e822

src/app/flows/fines/fines-reports/stores/interfaces/fines-reports-state.interface.ts
TimDanielsCQI: See above

src/app/flows/fines/fines-reports/fines-reports-select-business-units/fines-reports-select-business-units.component.html
TimDanielsCQI: Looks like it, the search form (similar functionality) does this. I have made the change.

Commit: https://github.com/hmcts/opal-frontend/pull/2680/changes/6e783b0a6096b09ac98d817297e5e45c30fba4dd

src/app/flows/fines/fines-reports/fines-reports-select-business-units/fines-reports-select-business-units-form/fines-reports-select-business-units-form.component.html
TimDanielsCQI: I was trying to deal with the situation where no business units are returned i.e. what to display. But it is not in the TDIA or the Jira Ticket, and if there are no business units assigned to a user I guess this is an error and will be dealt with by the common error interceptor anyway. I have removed.

Commit: https://github.com/hmcts/opal-frontend/pull/2680/changes/b2b14a4bf32c0273a3352d95c7472daca3d8b28d

src/app/flows/fines/fines-reports/constants/find-fines-reports-definition.constant.ts
TimDanielsCQI: This was removed by commit: https://github.com/hmcts/opal-frontend/pull/2680/changes/656e2b943cbe6ba677df6def25038a773f2129e6

src/app/flows/fines/constants/reports-primary-navigation-permissions.constant.ts
TimDanielsCQI: Sorry missed this in an earlier refactor. Have put it back to how it was.

Commit: https://github.com/hmcts/opal-frontend/pull/2680/changes/b5094bd3284d3da2beb57366147dbee345d98796

src/app/flows/fines/fines-reports/constants/fines-reports-operational-report-links.constant.ts
TimDanielsCQI: The backend changed the permission recorded against both operational reports to SEARCH_AND_VIEW_ACCOUNTS (PO-7274). This branch now uses the permission returned by GET /reports/{id} so the route guard and business-unit filtering match the backend report metadata.

src/app/flows/fines/fines-reports/fines-reports-select-business-units/fines-reports-select-business-units-form/fines-reports-select-business-units-form.component.ts
TimDanielsCQI: Zero and one business-unit results were incorrectly treated as the same valid case. I will restrict the read-only bypass to exactly one result and retain the required validator for an empty result, with a regression test confirming Continue cannot submit an empty selection.

Commit: https://github.com/hmcts/opal-frontend/pull/2680/changes/a56e4f3160a7105e05f09d87d9b28552b3045f77

src/app/flows/fines/fines-reports/fines-reports-business-unit-warning/fines-reports-business-unit-warning.component.ts
```diff
@@ -0,0 +1,51 @@
...
+  /**
+   * Redirects back to selection when no selected business unit ids are available in the reports store.
+   */
+  public ngOnInit(): void {
+    if (!this.finesReportsStore.hasSelectedBusinessUnits()) {
+      this.router.navigate([`../${FINES_REPORTS_CREATE_ROUTING_PATHS.children.selectBusinessUnits}`], {
+        relativeTo: this.activatedRoute,
+      });
+    }
+  }
```
iamfrankiemoran: This redirect is redundant because the warning route already sets requiresSelectedBusinessUnits: true, and finesReportsStateGuard redirects to business-unit selection before activating the component. Could we remove OnInit and ngOnInit() from this component?

src/app/flows/fines/fines-reports/fines-reports-select-business-units/fines-reports-select-business-units-form/fines-reports-select-business-units-form.component.ts
```diff
@@ -0,0 +1,218 @@
...
+    GovukCheckboxesItemComponent,
+    GovukCancelLinkComponent,
+    GovukErrorSummaryComponent,
+    GovukHeadingWithCaptionComponent,
+    GovukButtonDirective,
+  ],
+  templateUrl: './fines-reports-select-business-units-form.component.html',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class FinesReportsSelectBusinessUnitsFormComponent extends AbstractFormBaseComponent implements OnInit {
```
iamfrankiemoran, line 45: This implements the same underlying FormRecord and select-all behaviour as FinesSaSearchFilterBusinessUnitForm, but the two implementations now build and synchronise their controls differently. Could we align the report form with the established Fines SA structure—particularly createBusinessUnitControlsRecord(), buildBusinessUnitForm(), retrieving controls from the record, restoring selections and wiring select-all changes—while retaining the report-specific single-business-unit behaviour? Keeping two different implementations of the same selection mechanics will make future fixes harder to apply consistently.

src/app/flows/fines/fines-reports/fines-reports-select-business-units/fines-reports-select-business-units-form/fines-reports-select-business-units-form.component.html
```diff
@@ -0,0 +1,61 @@
...
+  (errorClick)="scrollTo($event)"
+></opal-lib-govuk-error-summary>
+
+<opal-lib-govuk-heading-with-caption
+  captionText="Create report"
+  [headingText]="reportHeading"
+></opal-lib-govuk-heading-with-caption>
+<h2 class="govuk-heading-m">{{ pageHeading }}</h2>
+
+<form (submit)="handleFormSubmit($event)" [formGroup]="form">
```
iamfrankiemoran: New bit of functionality incoming you probably weren't aware of

```suggestion
<form autocomplete="off" (submit)="handleFormSubmit($event)" [formGroup]="form">
```

src/app/flows/fines/fines-reports/fines-reports-business-unit-warning/fines-reports-business-unit-warning.component.ts
```diff
@@ -0,0 +1,51 @@
...
+import { FinesReportsStore } from '../stores/fines-reports.store';
+import { FINES_REPORTS_BUSINESS_UNIT_WARNING_CONTENT } from './constants/fines-reports-business-unit-warning-content.constant';
+
+@Component({
+  selector: 'app-fines-reports-business-unit-warning',
+  imports: [GovukButtonDirective, GovukCancelLinkComponent],
+  templateUrl: './fines-reports-business-unit-warning.component.html',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class FinesReportsBusinessUnitWarningComponent implements OnInit {
```
iamfrankiemoran: Could we follow the existing Fines ACC presentational-component pattern here? The static heading and description should live in the component template, with only selectedBusinessUnitCount supplied as dynamic data. This would remove the content constant, the content object and the heading function.

src/app/flows/fines/fines-reports/constants/fines-reports-operational-report-links.constant.ts
```diff
@@ -15,13 +16,13 @@ export const FINES_REPORTS_OPERATIONAL_REPORT_LINKS: IDashboardPageConfiguration
       FINES_ROUTING_PATHS.children.reports.children['summaryList'],
     ],
     fragment: null,
-    permissionIds: [FINES_PERMISSIONS['operational-report-by-enforcement']],
```
iamfrankiemoran, line 18: [P1]: Use report metadata as the permission source
Problem: These links and the corresponding route configuration statically require search-and-view-accounts, even though the required permission is supplied dynamically by GET /reports/{id}.
Why: Dashboard links are filtered before the report resolver runs, so a user who can access Reports but does not have this hard-coded permission cannot select the report. The current route guard also checks the static configuration rather than the permission returned by the endpoint.
Fix: Leave the dashboard link permissionIds empty and make the report metadata authoritative after navigation. The route configuration and resolver logic will also need to be decoupled from permissionIds.length: currently an empty array causes fetchReportResolver to skip GET /reports/{id} and causes finesReportsStateGuard to treat create routes as inaccessible. Access and business-unit filtering should instead use the permission returned for the selected report.
Example: Use permissionIds: [] for link visibility, identify operational reports independently of that array, and fetch/check report.permission when entering the report journey.

src/app/flows/fines/fines-reports/routing/fines-reports.routes.ts
```diff
@@ -29,8 +37,47 @@ export const routing: Routes = [
...
+            canDeactivate: [canDeactivateGuard],
+            data: {
+              title: FINES_REPORTS_ROUTING_TITLES.children.selectBusinessUnits,
+              requiresCreateReport: true,
+            },
+            resolve: {
+              title: TitleResolver,
+              report: fetchReportResolver,
+              reportHeading: finesReportsReportHeadingResolver,
+              businessUnits: fetchBusinessUnitsResolver,
```
iamfrankiemoran, line 61: [P1]: Prevent access when no business units are available
Problem: fetchBusinessUnitsResolver can successfully return an empty refData array, but the route still activates the business-unit selection screen.
Why: The form then contains no selectable controls while Continue remains visible, leaving the user in an impossible validation state. An empty successful response will not be handled by the HTTP error interceptor.
Fix: Use a report-specific resolver that redirects when the permission-filtered business-unit response is empty, while returning the response normally when at least one unit is available. Add a resolver test for the empty response.
Example: Return an Angular RedirectCommand, following the existing Fines ACC resolver redirect pattern.

iamfrankiemoran (changes requested):

src/app/flows/fines/fines-reports/constants/fines-reports-operational-report-links.constant.ts
TimDanielsCQI: The operational-report links and route configuration no longer hard-code `SEARCH_AND_VIEW_ACCOUNTS`. They now only identify whether a report requires metadata.

When the user enters an operational report journey, the frontend calls `GET /reports/{reportTypeId}` and uses the `permission` returned in that response. It then calls the business-unit endpoint with that permission to determine whether the user has access to any eligible business units.

This means the current API response will continue to work as-is, but when the backend is updated to return the report-specific permission, such as `OPERATIONAL_REPORT_BY_ENFORCEMENT`, the frontend will automatically use it without another frontend change. 

In other words it is dynamic now, so when DB team  change the specific permission to check, our code will not need to change.

Commit: https://github.com/hmcts/opal-frontend/pull/2680/changes/618b8f552837a46ad3f27a0222ed3caf9eba2ee2

src/app/flows/fines/fines-reports/fines-reports-business-unit-warning/fines-reports-business-unit-warning.component.ts
TimDanielsCQI: Yes it is redundant, sorry did not spot that after a refactor. it is gone now

Commit: https://github.com/hmcts/opal-frontend/pull/2680/changes/423e2ce9e298058e0437a56af8badbf96e0a58c5

src/app/flows/fines/fines-reports/routing/fines-reports.routes.ts
TimDanielsCQI: This is already covered by finesReportsStateGuard. It runs before the resolver and redirects to access denied when no business units are returned, so the selection page cannot load with an empty list.

See commit: https://github.com/hmcts/opal-frontend/pull/2680/changes/618b8f552837a46ad3f27a0222ed3caf9eba2ee2

src/app/flows/fines/fines-reports/fines-reports-select-business-units/fines-reports-select-business-units-form/fines-reports-select-business-units-form.component.ts
TimDanielsCQI: The reports form now uses createBusinessUnitControlsRecord(), buildBusinessUnitForm(), restoration from the control record, and separate subscriptions for individual and Select all changes.

I have not changed the Fines SA component, while making the equivalent reports selection mechanics easier to compare and maintain. Let me know if you think this refactor could go further, I was just trying to avoid changing the **FinesSaSearchFilterBusinessUnitForm**.

Commit: https://github.com/hmcts/opal-frontend/pull/2680/changes/04dcac01b5d3859f87e502e5f064097966ae5020

src/app/flows/fines/fines-reports/fines-reports-business-unit-warning/fines-reports-business-unit-warning-content/fines-reports-business-unit-warning-content.component.ts
```diff
@@ -0,0 +1,29 @@
...
+import { GovukCancelLinkComponent } from '@hmcts/opal-frontend-common/components/govuk/govuk-cancel-link';
+import { GovukButtonDirective } from '@hmcts/opal-frontend-common/directives/govuk-button';
+
+@Component({
+  selector: 'app-fines-reports-business-unit-warning-content',
+  imports: [GovukButtonDirective, GovukCancelLinkComponent],
+  templateUrl: './fines-reports-business-unit-warning-content.component.html',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class FinesReportsBusinessUnitWarningContentComponent {
```
iamfrankiemoran, line 11: [P2]: Avoid an unnecessary component boundary
Problem: This child component has no independent state or reusable behaviour; it only receives the selected count and proxies Continue and Go back events to the routed component.
Why: The split adds another component class, template, event API and test surface without simplifying the already-small warning component.
Fix: Keep the static content directly in `fines-reports-business-unit-warning.component.html`, retain `selectedBusinessUnitCount` and the navigation handlers in the routed component, and remove this child component. Looking at the resulting implementation, my earlier presentational-component wording over-specified the desired split.

src/app/flows/fines/fines-reports/routing/fines-reports.routes.ts
```diff
@@ -13,8 +20,9 @@ export const routing: Routes = [
     pathMatch: 'full',
   },
   {
-    path: ':reportId',
-    canActivate: [authGuard, finesReportsStateGuard],
+    path: ':reportTypeId',
```
iamfrankiemoran, line 23: [P2]: Separate Your reports from metadata-backed report routes
Problem: This dynamic route handles both the special `Your reports` journey and API-backed operational report types, which requires `requiresReportMetadata` to be added to otherwise presentational report configuration.
Why: `Your reports` is intended to be a separate collection page, whereas operational report routes must always resolve `GET /reports/{id}` for their permissions and metadata. Combining the journeys also makes operational child routes, such as create report, structurally available beneath `Your reports`.
Fix: Add an explicit `Your reports` route before `:reportTypeId`, containing only its summary-list journey, and reserve the dynamic route and its children for API-backed report types. The operational resolver and guard can then always fetch report metadata and `requiresReportMetadata` can be removed.

iamfrankiemoran (changes requested):
Some more changes

PR #2719, opened 2026-06-16
PO-6443: Added docker files

Ben-Edwards-cgi:
### Jira link
https://tools.hmcts.net/jira/browse/PO-6443

### Change description
Added docker files to support local dev


docker-compose.base.yml
```diff
@@ -0,0 +1,66 @@
...
+      - FEATURES_REDIS_ENABLED=true
+      - SESSION_SECURE_ENABLED=true
+      - CSRF_SECURE_ENABLED=true
+      - FEATURES_HELMET_ENABLED=true
+      - FEATURES_LAUNCH_DARKLY_ENABLED=true
+      - FEATURES_LAUNCH_DARKLY_STREAM=false
+      - FEATURES_SSO_ENABLED=true
+      - SESSION_SAME_SITE=lax
+      - CSRF_SAME_SITE=lax
+      - FEATURE_FLAGS_OVERRIDE=true
```
iamfrankiemoran: Could you move `- FEATURE_FLAGS_OVERRIDE=true` to above `- RELEASE_1A_ENABLED=true` please? 

Its just the override flag is related to the flags which will be overridden

iamfrankiemoran: Also, should the `FEATURE_FLAGS_OVERRIDE` be true? We only want to do that for local development/testing it should come from Launch Darkly

iamfrankiemoran (changes requested):
Minor comment

docker-compose.base.yml
Ben-Edwards-cgi: These dockers files are exclusively used for local development and would not be used in any deployed environment

iamfrankiemoran (approved):

PR #2752, opened 2026-06-23
PO-2634

Arnabsubedi233:
### Jira link

See [https://tools.hmcts.net/jira/browse/PO-2634](https://tools.hmcts.net/jira/browse/PO-2634)

### Change description

- Added the history tab defendant transformation logic
### Testing done

- unit tested
- locally tested

  * [ ] Yes
  * [x] No

### Checklist


src/app/flows/fines/fines-acc/fines-acc-defendant-details/fines-acc-defendant-details-history-and-notes-tab/constants/fines-acc-defendant-details-history-and-notes-tab-history-item-keys.constant.ts
```diff
@@ -0,0 +1,8 @@
+export const FINES_ACC_DEFENDANT_DETAILS_HISTORY_AND_NOTES_TAB_HISTORY_ITEM_KEYS = [
+  'history_items',
+  'historyItems',
+  'history',
```
iamfrankiemoran: Why do we have three histories? 🤔

src/app/flows/fines/fines-acc/services/constants/fines-acc-history-and-notes-details-alias-path-prefixes.constant.ts
```diff
@@ -0,0 +1 @@
+export const FINES_ACC_HISTORY_AND_NOTES_DETAILS_ALIAS_PATH_PREFIXES = ['', 'details.', 'data.', 'payload.'];
```
iamfrankiemoran, line 3: Should we have an empty value in here? I also wonder if it is worth removing the `.` from the others?

src/app/flows/fines/fines-acc/services/constants/fines-acc-history-and-notes-details-defendant-account-record-type.constant.ts
```diff
@@ -0,0 +1 @@
+export const FINES_ACC_HISTORY_AND_NOTES_DETAILS_DEFENDANT_ACCOUNT_RECORD_TYPE = 'defendantaccount';
```
iamfrankiemoran: All lower case?

src/app/flows/fines/fines-acc/services/utils/fines-acc-payload-transform-history-and-notes.utils.ts
```diff
@@ -0,0 +1,723 @@
...
+ * @param instalmentPeriod - The instalment period display label.
+ * @param effectiveDate - The formatted effective date.
+ * @returns The instalments part or null.
+ */
+function instalmentsPart(
+  instalmentAmount: string | null,
+  instalmentPeriod: string | null,
+  effectiveDate: string | null,
+): IFinesAccHistoryAndNotesDetailsPart | null {
+  const instalmentValues = [instalmentAmount, instalmentPeriod, effectiveDate].filter(isPresentString);
```
TimDanielsCQI, line 357: Codex has flagged this, and I thought it might have a point:

Could we tighten this so `Instalments:` is only emitted when the history item actually has instalment data?

At the moment `effectiveDate` is included in `instalmentValues`, so an `In full` or `Paid` payment terms item with only an `effective_date` produces an extra instalments part, for example:

`In full | Due by: 10/11/2025 | Instalments: 10/11/2025`

For those cases the date is the due/paid date, not evidence of an instalment schedule, so the rendered Details text would be misleading. This looks like an AC1e/template-alignment issue rather than just display formatting.

I think `instalmentsPart()` should return `null` unless there is an instalment amount or period, while still including `effectiveDate` when an actual instalment schedule exists.

Suggested shape:

```ts
if (!instalmentAmount && !instalmentPeriod) {
  return null;
}

File: src/app/flows/fines/fines-acc/services/utils/fines-acc-payload-transform-history-and-notes.utils.spec.ts [hunk 5]
```diff
@@ -0,0 +1,739 @@
...
+      },
+      FINES_ACC_HISTORY_AND_NOTES_DETAILS_TRANSFORMERS,
+    );
+
+    expect(result).toEqual({
+      line1: [part(fragment('Suspense transfer')), part(fragment('Unmapped suspense transfer reason'))],
+      line2: null,
+    });
+  });
+});
```
TimDanielsCQI, line 759: Again, Codex is flagging up that no test exists to cover special characters:

Could we add a focused unit test for the Jira “nested special characters” note?

There is already coverage for nested enforcement details, but I could not see a case where nested backend values contain characters such as `&`, `<`, `>`, quotes, or apostrophes. The Jira explicitly says nested special characters from the backend should not break the transformation, and QA review for this ticket is expected to focus on the unit tests.

A useful test would put special characters into nested values such as `details.hearingCourt.court_name`, `caseNumber`, and `reason`, then assert that the resulting fragments preserve the text exactly.

For example, something like:

```ts
hearingCourt: {
  court_name: 'Brent & Harrow <Magistrates> "Court"',
},
caseNumber: "Case 'A' & B",
reason: `Defendant said "can't pay" & requested <review>`,

TimDanielsCQI (commented):
Codex flags up special characters not being tested in the unit tests

src/app/flows/fines/fines-acc/services/utils/fines-acc-payload-transform-history-and-notes.utils.ts
Arnabsubedi233: Yep that is a real issue, thanks for that

TimDanielsCQI (approved):
LGTM

louisbriggs (approved):
LGTM

iamfrankiemoran (changes requested):
Three original comments

src/app/flows/fines/fines-acc/services/constants/fines-acc-history-and-notes-details-alias-path-prefixes.constant.ts
Arnabsubedi233: With the current common helper, both are required.
The helper does this for unqualified aliases:
aliasPathPrefixes.map((prefix) => `${prefix}${alias}`)
So:
'' + 'type' gives top-level type
'details.' + 'noteText' gives details.noteText
if we changed to ['details', 'data', 'payload'], it would look for detailsnoteText, datanoteText, etc.
if we removed '', top-level aliases like type would stop resolving

iamfrankiemoran (approved):
LGTM

Alfred-Gillingham2 (approved):

PR #2835, opened 2026-07-09
Apply safe retrieval retry policy to frontend read operations, using common UI retry interceptor

stantb:
### Jira link

https://tools.hmcts.net/jira/browse/PO-3937

### Change description

Upgrades @hmcts/opal-frontend-common to 0.0.90
Registers the shared httpRetryInterceptor in app.config.ts
Opts safe OpalFines read requests into retry using the common UI retry context helper
Keeps mutations and If-Match versioned requests out of retry

### Testing done

Covered new functionality with unit tests

  * [ ] Yes
  * [x] No

### Checklist


src/app/flows/fines/services/opal-fines-service/opal-fines.service.ts
iamfrankiemoran: [P2] Add retry context to minor creditor history reads

This PR opts safe OpalFines GETs into `withHttpRetry`, including the defendant history request, but the analogous minor creditor history request still builds plain `{ observe: 'response' }` options. A transient 408/504 on the minor creditor History and notes tab will still go straight to the error path while the defendant tab retries; build these options with `withRetrySafeReadOptions` too.

iamfrankiemoran (changes requested):

iamfrankiemoran (approved):
LGTM

src/app/flows/fines/services/opal-fines-service/opal-fines.service.ts
```diff
@@ -79,6 +80,19 @@ export class OpalFines {
   private readonly PARAM_ACCOUNT_STATUS_DATE_FROM = 'account_status_date_from';
   private readonly PARAM_ACCOUNT_STATUS_DATE_TO = 'account_status_date_to';
 
+  private retrySafeReadOptions() {
+    return { context: withHttpRetry({ retryCount: 1, delayMs: 0 }) };
```
louisbriggs: Should this config, passed to withHttpRetry, be stored in a const?

louisbriggs (commented):
Minor question

src/app/flows/fines/services/opal-fines-service/opal-fines.service.ts
stantb: Yes that makes sense, have made the change

louisbriggs (approved):
LGTM

TimGonella (dismissed):
A small gap still looks open here: the shared retry opt-in has been applied to most safe read-only OpalFines GETs, but the major creditor account reads still appear to be missing it in getMajorCreditorAccountHeadingData() and getMajorCreditorAccountAtAGlance().
Those two calls look in scope for this change for the same reason as the defendant/minor creditor reads: they’re safe retrieval requests, not mutations, and they don’t rely on If-Match write semantics. Without the retry context, transient 502/503/504/408 failures on major creditor account pages won’t get the same retry behaviour as the other account detail journeys

stephenmagill (approved):

TimGonella (approved):

PR #2836, opened 2026-07-09
PO-2462 - remove unrequired values from add, update and replace draft-accounts request payload

GarethLangley1:
### Jira link

See [PO-2462](https://tools.hmcts.net/jira/browse/PO-2462)

### Change description

- Removed the deprecated draft-account request fields from the FE mutation payloads so add, replace, and update/delete requests no longer send `submitted_by`, `submitted_by_name`, `validated_by`, or `validated_by_name`.
- Updated the draft-account request types and payload builders to reflect the new API contract, including the add/replace payload builder and the patch/delete payload builder.

### Testing done

- Updated the related request mocks so the test fixtures match the new request shape.
- Refactored the unit test helpers to build expected request payloads without relying on deleted fields.
Updated the affected unit tests to assert the new request shape and verified the focused specs still pass.

  * [ ] Yes
  * [x] No

### Checklist


iamfrankiemoran (approved):
LGTM

bpires2 (approved):
LGTM

stephenmagill (approved):

CadeFaulkner (changes requested):
A hotfix was put into the backend to allow additional json fields to be sent in the request, this will eventually be removed, therefore we need to ensure that all the fixtures are updated to remove the non-required fields, any steps that we have need to be updated to remove any code that is adding in these fields with values, for example submitted_by and submitted_by_name were previously required and as a result for our data setup we were ensuring this was provided with the details of the signed in user, the BE now works this out, and should reject payloads that are sent with it populated.

This is the BE hotfix, which will eventually need to be reverted, the FE and Tests will need to work as if this hotfix was not implemented. https://github.com/hmcts/opal-fines-service/pull/2249 

This needs to be tested against a branch that has the hotfix removed.

PR #2844, opened 2026-07-13
PO-8911-arent-always-taken-to-top-of-screen

marcmollins:
updated the application so whenever entering a screen, the user is always taken to the top of the screen.

### Jira link

https://tools.hmcts.net/jira/browse/PO-8911

### Change description

When a user scrolls down on a screen, enters another page and then goes back to the screen, they are expecting to go to the top, this now makes sure wherever on the application this occurs, we are sending the user to the top of the screen.

### Testing done

unit tests, cypress tests and tested locally

  * [ ] Yes
  * [x] No

### Checklist


louisbriggs (approved):
LGTM

src/app/app.config.ts
```diff
@@ -20,6 +20,10 @@ export const appConfig: ApplicationConfig = {
       withRouterConfig({
         canceledNavigationResolution: 'computed',
       }),
+      withInMemoryScrolling({
+        anchorScrolling: 'enabled',
+        scrollPositionRestoration: 'top',
+      }),
```
iamfrankiemoran, line 26: [P2] Keep fragment navigations scrolling to top

Enabling anchorScrolling makes Angular prefer matching #fragment anchors over the top reset on normal navigations. The payment-terms amend save/cancel paths navigate back with fragment 'payment-terms', and the details tab renders a matching id, so users land on the payment terms section rather than the top of the page. Remove this option or leave anchor scrolling disabled; scrollPositionRestoration: 'top' is enough to reset fragment navigations to the top.

iamfrankiemoran (changes requested):
@louisbriggs just so you have sight of this one

src/app/app.config.ts
marcmollins: Thanks Frankie - I have disabled AnchorScrolling in this PR

src/app/app.config.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+            scrollPositionRestoration: string;
+          };
+        },
+    );
+
+    expect(routerScroller.options).toMatchObject({
+      anchorScrolling: 'enabled',
+      scrollPositionRestoration: 'top',
+    });
+  });
```
iamfrankiemoran, line 84: [P1] Align the spec with disabled anchor scrolling

The implementation now sets anchorScrolling to 'disabled', but this new spec still expects 'enabled', so the focused app config test fails and will block the test suite. Update this expectation to 'disabled' to match the corrected router config.

iamfrankiemoran (changes requested):

src/app/app.config.ts
louisbriggs: I think this may be ok at the moment as we don't have matching id's in the tab contents? Although if an in-page id matches a fragment there may be undesired results in the future.

The main risk in this app is that fragments appear to be used for tab state, for example #at-a-glance, #payment-terms, etc. If the page contains elements with matching IDs, enabling anchorScrolling may cause the page to jump when switching tabs or loading a tabbed route. If there is no matching element ID, Angular will not visibly scroll to an anchor, but scrollPositionRestoration: 'top' may still move the page to the top on route navigation.

iamfrankiemoran (approved):
LGTM

iamfrankiemoran (changes requested):
PO-8911 should own this centrally. In the installed Angular 21 router, normal navigations with no fragment fall through to `scrollToPosition([0, 0])` when `scrollPositionRestoration` is not disabled, while navigations with a fragment call `scrollToAnchor(fragment)` when `anchorScrolling` is enabled. That matches the clarified requirement.

Please change the router config to:
```
withInMemoryScrolling({
  anchorScrolling: 'enabled',
  scrollPositionRestoration: 'top',
})
```
and update `app.config.spec.ts` accordingly. It would also be worth adding focused coverage for both branches: no-fragment navigation scrolls to top, and fragment navigation scrolls to the matching anchor, for example `party-details`.

I checked current fragment usage and enabling anchor scrolling globally can affect existing fragment routes with matching IDs, including account enquiry tab/detail fragments such as `payment-terms`, `defendant`, `individuals`, and `companies`. That looks consistent with the clarified behaviour, but it should be called out as the expected global behaviour change.

Once PO-8911 is merged with this central behaviour, PO-2671 should keep its fragment bindings and target IDs, but remove the manual `DOCUMENT` / `ngAfterViewInit` / `scrollIntoView` implementations.

@louisbriggs do you agree? 
@GarethLangley1 this relates to your PR you have pushed: https://github.com/hmcts/opal-frontend/pull/2857

louisbriggs:
> PO-8911 should own this centrally. In the installed Angular 21 router, normal navigations with no fragment fall through to `scrollToPosition([0, 0])` when `scrollPositionRestoration` is not disabled, while navigations with a fragment call `scrollToAnchor(fragment)` when `anchorScrolling` is enabled. That matches the clarified requirement.
> 
> Please change the router config to:
> 
> ```
> withInMemoryScrolling({
>   anchorScrolling: 'enabled',
>   scrollPositionRestoration: 'top',
> })
> ```
> 
> and update `app.config.spec.ts` accordingly. It would also be worth adding focused coverage for both branches: no-fragment navigation scrolls to top, and fragment navigation scrolls to the matching anchor, for example `party-details`.
> 
> I checked current fragment usage and enabling anchor scrolling globally can affect existing fragment routes with matching IDs, including account enquiry tab/detail fragments such as `payment-terms`, `defendant`, `individuals`, and `companies`. That looks consistent with the clarified behaviour, but it should be called out as the expected global behaviour change.
> 
> Once PO-8911 is merged with this central behaviour, PO-2671 should keep its fragment bindings and target IDs, but remove the manual `DOCUMENT` / `ngAfterViewInit` / `scrollIntoView` implementations.
> 
> @louisbriggs do you agree? @GarethLangley1 this relates to your PR you have pushed: #2857

This sounds like a good approach to me.

CadeFaulkner (approved):

src/app/app.config.ts
```diff
@@ -20,6 +20,10 @@ export const appConfig: ApplicationConfig = {
       withRouterConfig({
         canceledNavigationResolution: 'computed',
       }),
+      withInMemoryScrolling({
+        anchorScrolling: 'disabled',
```
iamfrankiemoran:
```suggestion
        anchorScrolling: 'enabled',
```

src/app/app.config.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+        routerScrollerProvider.useFactory?.() as {
+          options: {
+            anchorScrolling: string;
+            scrollPositionRestoration: string;
+          };
+        },
+    );
+
+    expect(routerScroller.options).toMatchObject({
+      anchorScrolling: 'disabled',
```
iamfrankiemoran:
```suggestion
      anchorScrolling: 'enabled',
```

iamfrankiemoran (changes requested):

iamfrankiemoran (approved):
LGTM ✅

PR #2852, opened 2026-07-14
PO-8902-Route-guard-does-not-trigger

marcmollins:
### Jira link

https://tools.hmcts.net/jira/browse/PO-8902

### Change description

make sure route guard shows when cancelling and changes have been made to the form in question

### Testing done

Unit tested, tested locally.

  * [ ] Yes
  * [x] No

### Checklist


iamfrankiemoran (changes requested):
The change needs to be made where the unsaved state is lost, rather than in the route guard.

1. Revert the guard change
In `fines-mac-offence-details.component.ts`, restore `canDeactivate()` so it only uses the existing shared `unsavedChanges` state:
```
canDeactivate(): CanDeactivateTypes {
  return !this.finesMacStore.unsavedChanges();
}
```

Please remove this check:
```
const hasDraftOffenceChanges = this.finesMacOffenceDetailsStore.offenceDetailsDraft().length > 0;
```

`offenceDetailsDraft` is temporary navigation state, not a dirty-state indicator. It remains populated after the offence is copied into the main MAC store, so using it here can show a warning after a successful save.

2. Restore the unsaved state after saving the minor creditor
In `fines-mac-offence-details-minor-creditor.component.ts`, update `handleMinorCreditorFormSubmit()` immediately after the creditor is written to the draft:
```
this.finesMacOffenceDetailsStore.setOffenceDetailsDraft(offenceDetailsDraft);
this.finesMacOffenceDetailsStore.setMinorCreditorAdded(true);
this.finesMacStore.setUnsavedChanges(true);

this.routerNavigate(FINES_MAC_OFFENCE_DETAILS_ROUTING_PATHS.children.addOffence);
```

The important new line is:
```
this.finesMacStore.setUnsavedChanges(true);
```

Please set the store directly here. Do not call:
```
this.handleUnsavedChanges(true);
```

That would also set `stateUnsavedChanges` on the minor-creditor component itself. Because that child route has its own `canDeactivate` guard, it could cause the warning to appear while navigating away after pressing Save.
```
The expected state lifecycle should be:
Minor-creditor form submitted
    → creditor stored in offence draft
    → finesMacStore.unsavedChanges = true

User clicks Cancel on Add an offence
    → outer route guard sees true
    → warning displayed

User saves/reviews the complete offence
    → the offence form’s normal submission emits false
    → finesMacStore.unsavedChanges = false
    → leaving Offence details does not warn
```

3. Move the regression test
Remove the new test from `fines-mac-offence-details.component.spec.ts`:
```
it('should block deactivation when offence draft data exists', ...)
```

Also remove its now-unused `FINES_MAC_OFFENCE_DETAILS_DRAFT_STATE_MOCK` import.
Instead, update `fines-mac-offence-details-minor-creditor.component.spec.ts`. In the existing submission test, after:
```
component.handleMinorCreditorFormSubmit(formSubmit);
```

add:
```
expect(finesMacStore.unsavedChanges()).toBe(true);
```

We should also have journey coverage for both outcomes:
```
Save minor-creditor details → return to Add an offence → Cancel → warning appears.
Save minor-creditor details → save/review the complete offence → leave Offence details → no warning appears.
```
This fixes the point where the state becomes incorrect and keeps the route guard concerned only with actual unsaved state.

src/app/flows/fines/fines-mac/fines-mac-offence-details/fines-mac-offence-details.component.ts
```diff
@@ -20,11 +20,7 @@ export class FinesMacOffenceDetailsComponent implements OnDestroy {
    * @returns A boolean indicating whether the component can be deactivated.
    */
   canDeactivate(): CanDeactivateTypes {
-    if (this.finesMacStore.unsavedChanges()) {
-      return false;
-    } else {
-      return true;
-    }
+    return !this.finesMacStore.unsavedChanges();
```
iamfrankiemoran: The change to `canDeactivate()` is behaviourally equivalent to the existing implementation and is unrelated to the route-guard fix. Could we revert this file to keep the ticket focused and minimise the diff? The required behavioural change is already handled by setting `finesMacStore.unsavedChanges(true)` after saving the minor creditor.

iamfrankiemoran (changes requested):
Minor comment

src/app/flows/fines/fines-mac/fines-mac-offence-details/fines-mac-offence-details.component.ts
marcmollins: i have removed that file from the diff

iamfrankiemoran (approved):
LGTM ✅ 
@marcmollins just run: 
`yarn install`
then `yarn audit:save`

louisbriggs (approved):
LGTM

iamfrankiemoran (commented):
Once minor-creditor details are saved into the offence draft, the offence as a whole should remain marked as having unsaved changes until the user saves/reviews the complete offence. `markAsDirty()` is therefore reasonable, but it should not be based on `minorCreditorAdded`, because that flag controls the success banner and is reset whenever the Add offence component is destroyed.
Could we introduce a persistent offence-level dirty flag in `FinesMacOffenceDetailsStore`, for example:
```
offenceDetailsDraftDirty: false
```

Set it when the minor creditor is saved:
```
this.finesMacOffenceDetailsStore.setOffenceDetailsDraftDirty(true);
```

Use it when recreating the Add offence form:
```
if (this.finesMacOffenceDetailsStore.offenceDetailsDraftDirty()) {
  this.form.markAsDirty();
}
```

Then clear it only after the complete offence has been successfully saved/reviewed:
```
this.finesMacOffenceDetailsStore.setOffenceDetailsDraftDirty(false);
```

It should also be cleared as part of the existing offence-details store reset when the user confirms that the draft can be discarded. This keeps the unsaved state across nested navigation and separates route-guard behaviour from the minor-creditor success-banner lifecycle.

@louisbriggs thoughts on the above?

cypress/e2e/functional/opal/actions/manual-account-creation/offence-review.actions.ts
```diff
@@ -213,6 +213,17 @@ export class ManualOffenceReviewActions {
...
+  /**
+   * Returns to Account details from review and asserts no unsaved-changes dialog appears.
+   */
+  returnToAccountDetailsWithoutUnsavedChangesWarning(): void {
+    log('navigate', 'Returning to account details from review without unsaved-changes warning');
+    const confirmSpy = cy.stub().returns(true);
+    cy.on('window:confirm', confirmSpy);
+    this.returnToAccountDetails();
+    cy.wrap(confirmSpy).should('not.have.been.called');
+  }
```
iamfrankiemoran, line 225: [P2]: E2E regression coverage. Problem: The new return-to-account-details-without-warning action is only referenced by its flow and step definition; the step is not used by any feature scenario. Why: The intended assertion that saving the offence clears the new dirty state is therefore never executed. Fix: Add the new step to the relevant saved-offence scenario, or remove the unused action, flow, and step if that behaviour is already covered elsewhere. Example: After reaching offence review, invoke 'When I return to account details from offence review without an unsaved changes warning'.

src/app/flows/fines/fines-mac/fines-mac-offence-details/fines-mac-offence-details-add-an-offence/fines-mac-offence-details-add-an-offence-form/fines-mac-offence-details-add-an-offence-form.component.spec.ts
```diff
@@ -680,6 +680,41 @@ describe('FinesMacOffenceDetailsAddAnOffenceFormComponent', () => {
...
+  it('should emit unsaved changes on cancel when returning from a saved minor creditor', () => {
+    finesMacOffenceDetailsStore.setOffenceDetailsDraftDirty(true);
+    finesMacOffenceDetailsStore.setEmptyOffences(true);
+    component['initialAddAnOffenceDetailsSetup']();
+    const unsavedChangesEmitSpy = vi.spyOn(component['unsavedChanges'], 'emit');
+
+    component.cancelLink();
+
+    expect(unsavedChangesEmitSpy).toHaveBeenCalledWith(true);
+  });
```
iamfrankiemoran, line 738: [P2]: Unit-test lifecycle. Problem: These tests change the store after fixture.detectChanges() has already run and then call initialAddAnOffenceDetailsSetup() manually. Why: This reconstructs the form and registers its listeners a second time, so the test does not represent the real Angular lifecycle and could hide subscription or initialisation problems. Fix: Create a fresh fixture, configure offenceDetailsDraftDirty before its first detectChanges(), and allow ngOnInit to initialise the component once. Example: set the store state, create the component, spy on unsavedChanges, call fixture.detectChanges(), then exercise cancelLink().

src/app/flows/fines/fines-mac/fines-mac-offence-details/fines-mac-offence-details-minor-creditor/fines-mac-offence-details-minor-creditor.component.ts
```diff
@@ -53,6 +53,8 @@ export class FinesMacOffenceDetailsMinorCreditorComponent
 
     this.finesMacOffenceDetailsStore.setOffenceDetailsDraft(offenceDetailsDraft);
     this.finesMacOffenceDetailsStore.setMinorCreditorAdded(true);
+    this.finesMacOffenceDetailsStore.setOffenceDetailsDraftDirty(true);
+    this.finesMacStore.setUnsavedChanges(true);
```
iamfrankiemoran: [P1]: Route-guard behaviour  
Problem: Every minor-creditor submission unconditionally marks the offence draft and global state as dirty, including when an existing creditor is saved without changes.  
Why: This can produce an unsaved-changes warning when nothing changed. Conversely, replacing the global state with only the creditor comparison could clear unsaved changes made elsewhere on the offence form.  
Fix: Before replacing the existing creditor data, determine whether the creditor is new or its submitted data has changed. Preserve both the existing offence-draft dirty state and the existing global unsaved state when calculating the values to store.  
Example: Opening an existing creditor, making no changes, saving it, and cancelling the offence should not warn—unless the offence already contained other unsaved changes.

iamfrankiemoran (changes requested):

src/app/flows/fines/fines-mac/fines-mac-offence-details/fines-mac-offence-details-minor-creditor/fines-mac-offence-details-minor-creditor.component.ts
marcmollins: have made the required changes

src/app/flows/fines/fines-mac/fines-mac-offence-details/fines-mac-offence-details-add-an-offence/fines-mac-offence-details-add-an-offence-form/fines-mac-offence-details-add-an-offence-form.component.spec.ts
marcmollins: have made the changes suggested

cypress/e2e/functional/opal/actions/manual-account-creation/offence-review.actions.ts
marcmollins: have made the changes

src/app/flows/fines/fines-mac/fines-mac-offence-details/fines-mac-offence-details-minor-creditor/fines-mac-offence-details-minor-creditor.component.ts
```diff
@@ -19,6 +19,17 @@ export class FinesMacOffenceDetailsMinorCreditorComponent
   private readonly finesMacOffenceDetailsStore = inject(FinesMacOffenceDetailsStore);
   private readonly finesMacStore = inject(FinesMacStore);
 
+  private creditorFormChanged(
```
louisbriggs, line 29: Can method docs be added here please?

louisbriggs (changes requested):
Minor change

src/app/flows/fines/fines-mac/fines-mac-offence-details/fines-mac-offence-details-minor-creditor/fines-mac-offence-details-minor-creditor.component.ts
```diff
@@ -32,16 +43,20 @@ export class FinesMacOffenceDetailsMinorCreditorComponent
     // Update the imposition position in the form data
     const removeMinorCreditor = this.finesMacOffenceDetailsStore.removeMinorCreditor();
     const offenceDetailsDraft = structuredClone(this.finesMacOffenceDetailsStore.offenceDetailsDraft());
+    const offenceDetailsDraftDirty = this.finesMacOffenceDetailsStore.offenceDetailsDraftDirty();
+    const unsavedChanges = this.finesMacStore.unsavedChanges();
```
iamfrankiemoran, line 54: [P1]: Route-guard behaviour
Problem: `unsavedChanges` is read after the minor-creditor form submission has synchronously emitted `false`, while `offenceDetailsDraftDirty` is not set when a dirty offence form enters the child route.
Why: Change an offence field, open an existing minor creditor, save it without changing it, then cancel the offence—the reconstructed form can be pristine and no warning appears, despite the parent change remaining only in the draft.
Fix: When `updateOffenceDetailsDraft()` captures the parent form before nested navigation, preserve `offenceDetailsDraftDirty() || hasUnsavedChanges()` in the store. The minor-creditor submit can then combine that persisted value with `hasCreditorChanges`. Add a test that starts with a dirty parent form rather than directly setting both stores to true.
Example: Dirty offence form → Change existing creditor → save creditor unchanged → Cancel offence must display the warning.

marcmollins: done

iamfrankiemoran (approved):

louisbriggs (changes requested):
P2 – Rehydrated dirty drafts do not sync the route guard state

fines-mac-offence-details-add-an-offence-form.component.ts (line 233) marks the form as dirty when offenceDetailsDraftDirty() is true, but it does not also update finesMacStore.unsavedChanges(true).

The top-level route guard reads this value from the store (fines-mac-offence-details.component.ts, line 22), so a restored dirty form can still leave the guard state as false if navigation doesn't go through the form's cancel handler.

Suggested fix:

Synchronise the global unsavedChanges state when hasSavedDraftChanges is applied.
Add a unit test to verify the store value is updated correctly.


P2 – Removal flows update the draft without marking it dirty

fines-mac-offence-details-remove-minor-creditor.component.ts (line 70) and fines-mac-offence-details-remove-imposition.component.ts (line 215) both update offenceDetailsDraft before navigating back, but neither sets offenceDetailsDraftDirty nor updates finesMacStore.unsavedChanges.

If the offence form was pristine before entering the removal confirmation flow, the removal creates an unsaved draft change that the route guard will not detect.

Suggested fix:

Set both offenceDetailsDraftDirty and finesMacStore.unsavedChanges(true) after a successful removal.
Add unit tests covering both the minor creditor and imposition removal flows to verify the route guard state is updated correctly.

louisbriggs (approved):
LGTM

TimGonella (approved):

PR #2913, opened 2026-07-24
PO-9113 - return to original page when account deletion is cancelled

sarahtw-cgi:
### Jira link

See https://tools.hmcts.net/jira/browse/PO-9113

### Change description

Navigate the user back to the Check Account Details screen when the user clicks 'Delete account' and then 'No - cancel' on the 'Check account details' page. 

### Testing done

- Updated existing unit tests and created new tests in `fines-mac-delete-account-confirmation.component.spec.ts` and `fines-mac-review-account.component.spec.ts`
- Manually followed the journey in the account create, account review, account validate journeys.

  * [ ] Yes
  * [x] No

### Checklist


iamfrankiemoran (changes requested):
[P1]: Angular correctness

Problem: `FinesMacDeleteAccountConfirmationComponent` reads `history.state` to determine where “No - cancel” should navigate, but this application must not use browser history to control flow navigation.

Why: Browser history is unavailable during SSR and makes the confirmation screen depend on browser state instead of the existing `FinesMacStore`. The required origin is already represented by `deleteFromCheckAccount`; it is just not set when deletion starts from the “Check account” screen without an `accountId`.

Fix:

1. Remove the `history.state` and `Router.currentNavigation()` lookup from `setReferrer()`.
2. Remove the `NavigationExtras.state`/`routeData` additions from the components that navigate to the confirmation screen.
3. In `FinesMacReviewAccountComponent.handleDeleteAccount()`, set `deleteFromCheckAccount` to `true` before checking `accountId`. It must be set for both new accounts without an ID and existing accounts with an ID.
4. Continue setting the flag to `true` when deletion starts from `FinesMacReviewAccountDecisionFormComponent`.
5. Leave `FinesMacAccountDetailsComponent` unchanged so the store flag remains `false` when deletion starts from “Account details”.
6. Retain the existing `checker()` check in `setReferrer()` and continue resetting `deleteFromCheckAccount` when the confirmation component is destroyed.

Example:

```ts
// FinesMacReviewAccountComponent
public handleDeleteAccount(event: Event, nonRelative = false): void {
  event.preventDefault();
  this.finesMacStore.setDeleteFromCheckAccount(true);

  const route =
    this.accountId > 0
      ? `${this.finesMacRoutes.children.deleteAccountConfirmation}/${this.accountId}`
      : this.finesMacRoutes.children.deleteAccountConfirmation;

  this.routerNavigate(route, nonRelative, event);
}
```

```ts
// FinesMacDeleteAccountConfirmationComponent
private setReferrer(): string {
  if (this.finesDraftStore.checker()) {
    return this.reviewAccountRoute;
  }

  const accountType =
    this.finesMacStore.accountDetails().formData.fm_create_account_account_type;

  return this.finesMacStore.deleteFromCheckAccount() ||
    accountType === FINES_ACCOUNT_TYPES['Fixed Penalty']
    ? this.reviewAccountRoute
    : this.accountDetailsRoute;
}
```

Please add tests proving all three journeys:

- Account creation → “Account details” → “Delete account” → “No - cancel” returns to “Account details”.
- Account creation → “Check account” → “Delete account” → “No - cancel” returns to “Check account”, including when there is no `accountId`.
- Check and validate → checker reviews an account → “Delete account” → “No - cancel” returns to the same “Check account” route with its `accountId`.

This keeps the check-and-validate behaviour unchanged while fixing the account-creation route using the existing store state.

iamfrankiemoran (approved):
LGTM ✅

ChrisGarratt122 (approved):
LGTM

TimGonella (approved):

PR #2914, opened 2026-07-24
PO-2656 add history and notes tab to major creditor

GarethLangley1:
### Jira link

See [PO-2656](https://tools.hmcts.net/jira/browse/PO-2656)

### Change description

- Added a new History and notes tab to the major creditor account details flow, following the existing tab structure used in account enquiry.
- Added the initial major creditor History and notes tab UI scaffold with the shared Show filter section, including Date from, Date to, and Filter, with no table rendering yet (this is to implemented under [PO-2658](https://tools.hmcts.net/jira/browse/PO-2658)).
- Reused the shared history-and-notes filter form and updated it so category options are only rendered when categories are actually supplied.
- Wired the major creditor tab into the frontend data flow so selecting the tab triggers GET `/major-creditor-accounts/{id}/history` for the current account.
- Added support for optional major creditor history date filters, with blank values omitted from the request and requests only sent when the user selects Filter.
- Implemented frontend conversion of submitted local dates into RFC3339 UTC timestamps before sending dateFrom and dateTo query params.
- Reused the shared date-range validation pattern so an error is shown when Date from is later than Date to.
- Added major creditor history service/cache plumbing, including the new cache entry and tab-to-cache mapping needed to support the tab’s initial unfiltered load pattern.
- Kept the major creditor feature layer date-only by removing category-specific UI/config from that tab, while leaving the lower-level API service contract aligned with the endpoint’s optional itemTypes support.

### Testing done

- Added and updated mocks/specs around the new tab, payload building, shared date conversion utility, and Opal fines service integration.

  * [ ] Yes
  * [x] No

### Checklist


src/app/flows/fines/fines-acc/services/utils/fines-acc-date-to-rfc3339-utc-timestamp.utils.ts
Arnabsubedi233, line 1: I dont think we have to convert to timestamp for the dates well atleast we should not do, because we are not capturing time ,only dates, so we should just be sending a date object not datetime. I think if you look at how fines-acc-defendant-history-and-notes transformation has been built, it should be similar to that when it comes to the dates.

Arnabsubedi233 (changes requested):
Just one small comment, the rest looks good

src/app/flows/fines/fines-acc/services/utils/fines-acc-date-to-rfc3339-utc-timestamp.utils.ts
iamfrankiemoran, line 1: I'm aware this wasn't your change, but I'm wondering if this could use the date service instead via the common-ui-lib 🤔 

> [P1]: Shared date handling
> Problem: This adds custom frontend date parsing and UTC timestamp conversion in `finesAccDateToRfc3339UtcTimestamp`.
> Why: We are trying to keep date parsing/formatting rules inside the shared date service/lib so timezone, validation, and formatting behaviour stays consistent across OPAL.
> Fix: Replace the manual `split('/')`, `Date.UTC`, and `toISOString()` logic with the shared `DateService`/lib date conversion helper, or add the missing helper there if it does not already exist.

Suggestion: 
```
const parsedDate = this.dateService.getFromFormat(date, 'dd/MM/yyyy');

if (!this.dateService.isValidDate(parsedDate)) {
  // handle invalid date consistently
}

return parsedDate
  .setZone('utc', { keepLocalTime: true })
  .startOf('day')
  .toISO();
```

src/app/flows/fines/fines-acc/fines-acc-major-creditor-details/fines-acc-major-creditor-details-history-and-notes-tab/interfaces/fines-acc-major-creditor-details-history-and-notes-filter-payload.interface.ts
```diff
@@ -0,0 +1,4 @@
+export interface IFinesAccMajorCreditorDetailsHistoryAndNotesFilterPayload {
+  dateFrom?: string;
+  dateTo?: string;
+}
```
iamfrankiemoran:
```suggestion
export interface IFinesAccMajorCreditorDetailsHistoryAndNotesFilterPayload {
  dateFrom: string | null;
  dateTo: string| null;
}
```

src/app/flows/fines/services/opal-fines-service/opal-fines.service.ts
```diff
@@ -1495,4 +1499,48 @@ export class OpalFines {
...
+   */
+  public getMajorCreditorAccountHistoryAndNotesTabData(
+    account_id: number | null,
+    filterParams?: IOpalFinesMajorCreditorAccountHistoryParams,
+  ): Observable<IOpalFinesAccountMajorCreditorDetailsHistoryAndNotesTabRefData> {
+    const url = `${OPAL_FINES_PATHS.majorCreditorAccounts}/${account_id}/history`;
+    const options: {
+      observe: 'response';
+      params?: Record<string, string>;
+    } = { observe: 'response' };
```
iamfrankiemoran: > [P2]: Safe read retry policy
> Problem: The new major-creditor history GET builds its HTTP options inline and does not use the existing safe-read retry wrapper.
> Why: Other account-detail GETs use `withRetrySafeReadOptions`, so this endpoint will behave differently on transient timeout/gateway failures.
> Fix: Initialise the options with `this.withRetrySafeReadOptions({ observe: 'response' as const })`, matching the minor-creditor history endpoint.

Could we extract the following into a private method and call it from these methods to reduce duplication? 
```
const options: {
      observe: 'response';
      params?: Record<string, string>;
    } = this.withRetrySafeReadOptions({ observe: 'response' as const });
```

iamfrankiemoran (changes requested):

src/app/flows/fines/fines-acc/services/utils/fines-acc-date-to-rfc3339-utc-timestamp.utils.ts
GarethLangley1: RFC3339 timestamps is part of the acceptance criteria (AC2b), however, after reading your comment I have reached out to a BA to seek confirmation and have been advised: "On reflection, I think I was over zealous with the ACs attempting to meet SYS-NFR-004. That NFR is specific about recording and displaying dates - which this search request is neither".

With the above in mind, I have removed all functionality and references to RFC3339 in this branch and converted to using `yyyy-MM-dd` format instead.

GarethLangley1: This is no longer relevant, please see comment: https://github.com/hmcts/opal-frontend/pull/2914#discussion_r3666473245

src/app/flows/fines/fines-acc/fines-acc-major-creditor-details/fines-acc-major-creditor-details-history-and-notes-tab/interfaces/fines-acc-major-creditor-details-history-and-notes-filter-payload.interface.ts
GarethLangley1: Updated 👍

src/app/flows/fines/services/opal-fines-service/opal-fines.service.ts
GarethLangley1: Updated 👍

src/app/flows/fines/fines-acc/fines-acc-major-creditor-details/fines-acc-major-creditor-details-history-and-notes-tab/interfaces/fines-acc-major-creditor-details-history-and-notes-filter-payload.interface.ts
iamfrankiemoran: @GarethLangley1 they are still optional?

iamfrankiemoran (changes requested):

src/app/flows/fines/fines-acc/fines-acc-major-creditor-details/fines-acc-major-creditor-details-history-and-notes-tab/interfaces/fines-acc-major-creditor-details-history-and-notes-filter-payload.interface.ts
GarethLangley1: @iamfrankiemoran  I've just cross-checked the API schema again for the major-creditor history filter and it defines `dateFrom` and `dateTo` as _optional_ query parameters, so the TS type should reflect omission rather than null, in order to match the schema contract.

Arnabsubedi233 (approved):
LGTM

iamfrankiemoran (approved):
LGTM ✅

Nipa1410 (approved):

PR #2969, opened 2026-08-04
PO-2963 only display 'Paid out' tile for minor creditor with no associated defendant

GarethLangley1:
### Jira link

See [PO-2963](https://tools.hmcts.net/jira/browse/PO-2963)

### Change description

- Updated the Minor Creditor header summary so repayment-style accounts now show only the Paid out tile, while Awarded, Awaiting payout, and Outstanding are hidden when there is no associated defendant.

### Testing done

- Added unit coverage for the minor creditor details page to verify the repayment view renders only Paid out and suppresses the other financial tiles.

  * [ ] Yes
  * [x] No

### Checklist


iamfrankiemoran (approved):
LGTM ✅

ChrisGarratt122 (approved):
LGTM

src/app/flows/fines/fines-acc/fines-acc-minor-creditor-details/fines-acc-minor-creditor-details.component.html
```diff
@@ -41,12 +41,14 @@ <h2 opal-lib-custom-account-information-item-label>Business Unit:</h2>
         {{ accountData.financials.paid_out | monetary }}
       </p>
     </opal-lib-custom-summary-metric-bar-item>
-    <opal-lib-custom-summary-metric-bar-item backgroundColour="govuk-lighter-blue-background-colour">
-      <p opal-lib-custom-summary-metric-bar-item-label textColour="govuk-dark-blue-text-colour">Awaiting payout:</p>
-      <p opal-lib-custom-summary-metric-bar-item-value textColour="govuk-dark-blue-text-colour">
-        {{ accountData.financials.awaiting_payout | monetary }}
-      </p>
-    </opal-lib-custom-summary-metric-bar-item>
+    @if (accountData.creditor.has_associated_defendant) {
```
louisbriggs, line 44: Should these two @if statements be combined into a single conditional wrapper?

louisbriggs (commented):
Minor comment

src/app/flows/fines/fines-acc/fines-acc-minor-creditor-details/fines-acc-minor-creditor-details.component.html
GarethLangley1: I did consider this, however, looking at the previous code, conditional statements have been set per tile (line:30 & line:52). I assume this is deliberate, to preempt any potential change, in the order in which the tiles are rendered.

TimGonella (changes requested):
[P1]: Angular Correctness
Problem: [fines-acc-minor-creditor-details.component.html (line 30)] now decides the header tiles purely from has_associated_defendant, but the ticket says the simplified summary applies to repayment accounts identified by the new repayment flag.
Why: This ignores the API contract in the story and will hide Awarded, Awaiting payout, and Outstanding for any non-repayment minor creditor that also lacks an associated defendant. The gap is reinforced by [fines-acc-minor-creditor-details-header.interface.ts (line 4)] which still has no repayment field at all.
Fix: Add creditor.repayment to the header interface/mocks and gate the simplified layout on the repayment condition, not just defendant association.
Example: Derive something like isRepaymentAccount = accountData.creditor.repayment && !accountData.creditor.has_associated_defendant and use that for the tile visibility rules.

Problem: The new tests do not actually lock in the repayment contract. The unit test at [fines-acc-minor-creditor-details.component.spec.ts (line 160)] only proves the “no associated defendant” case, and the Cypress spec at [AccountEnquiryHeader.cy.ts (line 833)] adds repayment: true via an untyped object even though the production interface does not contain that field.
Why: If the component keeps ignoring repayment, these tests still pass, so the branch can regress the intended story behaviour without the suite catching it.
Fix: Update the shared header type/mock to include repayment, then add coverage for both repayment: true and repayment: false when has_associated_defendant is false.

isaac-lewis2 (approved):
lgtm

PR #2984, opened 2026-08-06
PO-9139 - Add hidden captions value to impositions table

sarahtw-cgi:
### Jira link

https://tools.hmcts.net/jira/browse/PO-9139

### Change description

- Adds a visually hidden caption element so assistive technologies announce the title of the offence and the offence code for each table on the Offences and Impositions page

### Testing done
- Added a unit test to ensure that captions are rendered and visually hidden
- Manually tested:

Screenshot 2026-08-06 at 15 58 56
 
  * [ ] Yes
  * [x] No

### Checklist


iamfrankiemoran (changes requested):

iamfrankiemoran (changes requested):

src/app/flows/fines/fines-mac/fines-mac-offence-details/fines-mac-offence-details-review-offence/fines-mac-offence-details-review-offence.component.ts
```diff
@@ -25,6 +33,19 @@ export class FinesMacOffenceDetailsReviewOffenceComponent implements OnInit {
...
+    const offenceCode = this.offence.formData.fm_offence_details_offence_cjs_code!;
+    const offenceId = this.offence.formData.fm_offence_details_offence_id;
+
+    return this.opalFinesService.getOffenceByCjsCode(offenceCode).pipe(
+      map((response: IOpalFinesOffencesRefData) => {
+        const exactMatch = this.offenceDetailsService.findExactOffenceMatch(response, offenceCode, offenceId);
+        return exactMatch?.offence_title ?? response.refData[0]?.offence_title ?? '';
+      }),
+    );
+  }
```
iamfrankiemoran: [P1] Preserve the selected offence record

[P1]: Preserve the selected offence record. Problem: `getOffenceTitle()` discards the selected offence and returns only its title, while the caption takes its code from the form. The fallback can therefore combine data from different offences; the current test data produces `ak test (OFF123)` even though `ak test` belongs to `AK123456`. Why: The visually hidden caption can identify the imposition table as the wrong offence. Fix: Return the selected `IOpalFinesOffences` record and use both `offence_title` and `get_cjs_code` from that same record. Example: `map(response => this.offenceDetailsService.findExactOffenceMatch(response, offenceCode, offenceId) ?? response.refData[0] ?? null)`.

src/app/flows/fines/fines-mac/fines-mac-offence-details/fines-mac-offence-details-review-offence/fines-mac-offence-details-review-offence.component.spec.ts
```diff
@@ -91,4 +106,66 @@ describe('FinesMacOffenceDetailsReviewOffenceComponent', () => {
     expect(component.impositionRefData).toEqual(existingResults);
     expect(component.majorCreditorRefData).toEqual(existingMajorCreditors);
   });
+
+  describe('offenceTitle$', () => {
+    it('should expose the exact matched offence title', async () => {
```
iamfrankiemoran: [P2] Make the exact-match test exercise an exact match

[P2]: Test the intended matching branch. Problem: This test is named as an exact-match test, but the form contains `OFF123` while the response contains `AK123456`, so it actually passes through the first-result fallback. Why: The test passes while masking the title/code mismatch in the rendered caption. Fix: Supply matching offence code and ID values and assert the rendered caption contains both values from the selected record. Example: set the form code to `AK123456`, use offence ID `314441`, then assert `ak test (AK123456)`.

src/app/flows/fines/fines-mac/fines-mac-offence-details/fines-mac-offence-details-review-offence/fines-mac-offence-details-review-offence-heading/fines-mac-offence-details-review-offence-heading.component.ts
```diff
@@ -11,25 +8,15 @@ import { FinesMacOffenceDetailsReviewOffenceHeadingTitleComponent } from './fine
   templateUrl: './fines-mac-offence-details-review-offence-heading.component.html',
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
-export class FinesMacOffenceDetailsReviewOffenceHeadingComponent implements OnInit {
-  private readonly opalFinesService = inject(OpalFines);
-
+export class FinesMacOffenceDetailsReviewOffenceHeadingComponent {
   @Input({ required: true }) public offenceId!: number;
   @Input({ required: false }) public offenceRefId: number | null = null;
```
iamfrankiemoran: [P2] Remove the remaining unused offenceRefId input

[P2]: Remove the obsolete component input. Problem: `offenceRefId` is still declared and supplied by the parent, but the heading component no longer reads or forwards it. Why: This leaves a misleading input in the component API and suggests the reference ID affects rendering when it does not. Fix: Remove the `offenceRefId` input and its parent template binding. Example: delete this declaration and `[offenceRefId]="offence.formData.fm_offence_details_offence_id"`.

iamfrankiemoran (changes requested):

src/app/flows/fines/fines-mac/fines-mac-offence-details/fines-mac-offence-details-review-offence/fines-mac-offence-details-review-offence.component.html
```diff
@@ -1,20 +1,23 @@
...
-@if (showDetails) {
-  <app-fines-mac-offence-details-review-offence-imposition
-    [impositionRefData]="impositionRefData"
-    [majorCreditorRefData]="majorCreditorRefData"
-    [impositions]="offence.formData.fm_offence_details_impositions"
-    [offenceIndex]="offence.formData.fm_offence_details_id"
+@if (offenceDetails$ | async; as offenceDetails) {
+  <app-fines-mac-offence-details-review-offence-heading
+    [offenceId]="offence.formData.fm_offence_details_id"
+    [offenceCode]="offence.formData.fm_offence_details_offence_cjs_code!"
```
iamfrankiemoran: [P1] Use the selected offence code in the heading

[P1]: Keep offence title and code from the same record. Problem: The heading receives its title from `offenceDetails`, but still receives its code from the form. Why: In the supported fallback path, this creates inconsistent content—for example, the current test data displays title `ak test` with heading code `OFF123`, while the table caption uses `AK123456`. The heading actions also use the incorrect form code in their visually hidden text. Fix: Supply the heading code from the selected reference-data record. Example: `[offenceCode]="offenceDetails.get_cjs_code"`.

src/app/flows/fines/fines-mac/fines-mac-offence-details/fines-mac-offence-details-review-offence/fines-mac-offence-details-review-offence.component.ts
```diff
@@ -38,12 +64,23 @@ export class FinesMacOffenceDetailsReviewOffenceComponent implements OnInit {
...
 
+  /**
+   * Returns the formatted caption for the provided offence details.
+   *
+   * @param offenceDetails - The offence details used to generate the caption.
+   * @returns The formatted offence title and CJS code.
+   */
+  public getOffenceCaption(offenceDetails: IOpalFinesOffences): string {
+    return this.offenceDetailsService.getFormattedTitleAndCode(offenceDetails);
+  }
```
iamfrankiemoran: [P2] Remove unnecessary caption formatting methods

[P2]: Keep simple presentation formatting local. Problem: `getOffenceCaption()` delegates to a new service method solely to concatenate two fields already available in the template. Why: This introduces an unnecessary public service API, a method call during template evaluation, and several tests for trivial string interpolation. Fix: Remove `getOffenceCaption()` and `getFormattedTitleAndCode()` and bind the caption directly. Example: `[offenceCaption]="offenceDetails.offence_title + ' (' + offenceDetails.get_cjs_code + ')'"`.

src/app/flows/fines/fines-mac/fines-mac-offence-details/fines-mac-offence-details-review-offence/fines-mac-offence-details-review-offence-imposition/fines-mac-offence-details-review-offence-imposition.component.spec.ts
```diff
@@ -35,6 +37,7 @@ describe('FinesMacOffenceDetailsReviewOffenceImpositionComponent', () => {
     mockOpalFinesService = {
       getResultPrettyName: vi.fn().mockReturnValue(OPAL_FINES_RESULT_PRETTY_NAME_MOCK),
       getMajorCreditorPrettyName: vi.fn().mockReturnValue(OPAL_FINES_MAJOR_CREDITOR_PRETTY_NAME_MOCK),
+      getOffenceByCjsCode: vi.fn().mockReturnValue(of(OPAL_FINES_OFFENCES_REF_DATA_MOCK)),
```
iamfrankiemoran: [P2] Remove the obsolete offence lookup mock

[P2]: Keep the imposition test isolated from parent lookup behavior. Problem: The imposition spec still mocks `getOffenceByCjsCode` and imports offence reference data even though this component no longer performs that lookup. Why: The test setup misleadingly suggests that the child still owns an Opal Fines Service call. Fix: Remove the `getOffenceByCjsCode` mock and the now-unused offence-data and RxJS imports. Example: Keep only the `getResultPrettyName` and `getMajorCreditorPrettyName` methods required by this component.

iamfrankiemoran (changes requested):

src/app/flows/fines/fines-mac/fines-mac-offence-details/fines-mac-offence-details-review-offence/fines-mac-offence-details-review-offence.component.ts
sarahtw-cgi: I don't think we can bind this directly as suggested because get_cjs_code isn't available on offenceDetails.

There is a cjs_code property, but this is frequently undefined, so using:
[offenceCaption]="offenceDetails.offence_title + ' (' + offenceDetails.cjs_code + ')'"
would result in an incorrect/missing code being announced.

getFormattedTitleAndCode() resolves/calculates the correct CJS code before formatting the caption. That's why I introduced getOffenceCaption() to expose that resolved value to the template.

iamfrankiemoran: I’ve checked the type used by this specific lookup. `offenceDetails$` emits `IOpalFinesOffences`, selected from `IOpalFinesOffencesRefData.refData`, and `IOpalFinesOffences` defines `get_cjs_code` as a required string. The separate search-offence model is the one that uses `cjs_code`.
`getFormattedTitleAndCode()` currently resolves `offence.get_cjs_code ?? offence.cjs_code`, but `cjs_code` is not part of the declared `IOpalFinesOffences` response. The tests supporting that fallback have to cast incomplete objects to make them compile.
Therefore the direct binding is valid for the response type used here:
`[offenceCaption]="offenceDetails.offence_title + ' (' + offenceDetails.get_cjs_code + ')'"`
If the offence reference-data endpoint can genuinely return legacy records containing only `cjs_code`, could you point me to the API contract or an example payload? In that case, the interface should be corrected and the response normalised when it is mapped, rather than combining compatibility resolution and caption formatting in a public service method.

src/app/flows/fines/fines-mac/fines-mac-offence-details/fines-mac-offence-details-review-offence/fines-mac-offence-details-review-offence.component.html
```diff
@@ -1,20 +1,23 @@
...
-  (actionClicked)="emitAction($event)"
-></app-fines-mac-offence-details-review-offence-heading>
-
-@if (showDetails) {
-  <app-fines-mac-offence-details-review-offence-imposition
-    [impositionRefData]="impositionRefData"
-    [majorCreditorRefData]="majorCreditorRefData"
-    [impositions]="offence.formData.fm_offence_details_impositions"
-    [offenceIndex]="offence.formData.fm_offence_details_id"
+@if (offenceDetails$ | async; as offenceDetails) {
```
louisbriggs: [P1]: Empty Lookup Hides Saved Offence Content
Problem: wrap the whole offence heading and imposition table in @if (offenceDetails$ | async; as offenceDetails), but [getOffenceDetails() (line 47)] can return null when the ref-data lookup has no results.
Why: A saved/draft offence whose reference-data lookup returns an empty result will disappear from review/remove-offence screens, including its impositions and action controls. Before this change, the imposition table rendered independently of the offence title lookup.
Fix: Keep rendering the saved offence from form state, and only use the lookup result to enrich the title/caption. For example, map to a fallback object using the stored offence code and an empty or stored title, or split the template so the imposition table is not gated by the lookup result. Add a DOM test for the empty refData case.
Example:
map((response) => {
  const match = this.offenceDetailsService.findExactOffenceMatch(response, offenceCode, offenceId);
  return {
    cjs_code: match?.cjs_code ?? offenceCode,
    offence_title: match?.offence_title ?? '',
  };
})

louisbriggs (changes requested):
Change to handling of no results.

src/app/flows/fines/fines-mac/fines-mac-offence-details/fines-mac-offence-details-review-offence/fines-mac-offence-details-review-offence.component.ts
```diff
@@ -25,6 +34,24 @@ export class FinesMacOffenceDetailsReviewOffenceComponent implements OnInit {
...
+
+    return this.opalFinesService.getOffenceByCjsCode(offenceCode).pipe(
+      map((response: IOpalFinesOffencesRefData) => {
+        const match = this.offenceDetailsService.findExactOffenceMatch(response, offenceCode, offenceId);
+
+        return {
+          ...match,
+          cjs_code: match?.cjs_code ?? offenceCode,
+          offence_title: match?.offence_title ?? '',
+        } as IOpalFinesOffences;
```
iamfrankiemoran: [P2] Use a narrow display model

[P2]: Preserve accurate type contracts. Problem: The empty-response fallback contains only `cjs_code` and `offence_title`, but it is asserted as a complete `IOpalFinesOffences` object. Why: This hides the absence of required fields such as `offence_id` and `business_unit_id`, allowing future consumers to access data that does not exist. Fix: Type `offenceDetails$` and `getOffenceDetails()` as a narrow view model containing only the two fields consumed by the template. Example: `type OffenceDisplayDetails = Pick<IOpalFinesOffences, 'cjs_code' | 'offence_title'>;` and return the object without a type assertion.

iamfrankiemoran (changes requested):

iamfrankiemoran (approved):
LGTM ✅

louisbriggs (approved):
LGTM

CadeFaulkner (approved):

PR #3005, opened 2026-08-13
PO-8991 Adopt shared application initializer for trace propagation

stantb:
### Jira link

See [PROJ-8991](https://tools.hmcts.net/jira/browse/PROJ-8991)

### Change description

Updates OPAL Frontend to use provideOpalAppInitializer() from @hmcts/opal-frontend-common 0.0.105

The Application Insights browser SDK is loaded asynchronously. Previously, Angular did not wait for that initialization to complete, so early fetch requests could occur before the SDK had instrumented them. This could prevent `traceparent` headers being added and result in incomplete end-to-end transaction breakdowns in Application Insights.

### Testing done

- Started the app locally with Application Insights enabled.
- Confirmed browser API requests include a `traceparent` header.
- Confirmed Application Insights links the browser fetch request to the OPAL frontend server request.

  * [ ] Yes
  * [x] No

### Checklist


iamfrankiemoran (commented):
[P2]: Centralize the Angular initializer
Problem: This app-specific implementation would need to be duplicated in opal-rm-frontend, which uses the same asynchronous AppInitializerService.
Why: Duplicating the TransferState key, guard, and Promise handling risks the two frontends drifting and leaves RM with the same early-request tracing issue.
Fix: Move this behavior into an exported provideOpalAppInitializer() provider in opal-frontend-common-ui-lib, then consume that provider from both frontend repositories. This belongs in the common UI library rather than common-node-lib because it depends on Angular bootstrap and browser Application Insights.
Example: providers: [..., provideOpalAppInitializer()]

Could I get your thoughts on this @louisbriggs / @max-holland

louisbriggs:
> [P2]: Centralize the Angular initializer Problem: This app-specific implementation would need to be duplicated in opal-rm-frontend, which uses the same asynchronous AppInitializerService. Why: Duplicating the TransferState key, guard, and Promise handling risks the two frontends drifting and leaves RM with the same early-request tracing issue. Fix: Move this behavior into an exported provideOpalAppInitializer() provider in opal-frontend-common-ui-lib, then consume that provider from both frontend repositories. This belongs in the common UI library rather than common-node-lib because it depends on Angular bootstrap and browser Application Insights. Example: providers: [..., provideOpalAppInitializer()]
> 
> Could I get your thoughts on this @louisbriggs / @max-holland

It seems like a good idea to move this to the common ui lib.

Arnabsubedi233 (approved):
LGTM

src/app/app.config.ts
```diff
@@ -37,9 +37,6 @@ export const appConfig: ApplicationConfig = {
         cookieName: 'XSRF-TOKEN',
       }),
     ),
-    provideAppInitializer(async () => {
-      const appInitializerService = inject(AppInitializerService);
-      appInitializerService.initializeApp();
-    }),
+    provideOpalAppInitializer(),
```
iamfrankiemoran: [P1] Restore initializer integration coverage

[P1]: Testing Problem: The migration to provideOpalAppInitializer() removes the tests proving bootstrap waits for initialization and safely skips without server transfer state. Why: The ticket’s core Application Insights timing regression could return without failing this repository’s tests. Fix: Retain or adapt the ApplicationInitStatus tests from the first commit to exercise the common provider integration. Example: Mock AppInitializerService.initializeApp(), resolve its promise, and assert ApplicationInitStatus.done remains false until resolution.

iamfrankiemoran (changes requested):

iamfrankiemoran (approved):
LGTM ✅

src/app/app.config.ts
stantb: Have restored test coverage

iamfrankiemoran (approved):

louisbriggs (approved):
LGTM

PR #3030, opened 2026-08-24
PO-9041

louisbriggs:
### Jira link

See [PO-9041](https://tools.hmcts.net/jira/browse/PO-9041)

### Change description

- Updating ui lib package to introduce aria-live to banner alert messages.

- Improves accessibility of the session expiry warning by using a dedicated deferred live region announcement. Disables the MOJ alert's built-in live announcement to prevent duplicate/repeated announcements as the countdown updates.

### Testing done

Verified that alerts are being applied to banners as expected including the Expiry message

  * [ ] Yes
  * [x] No

### Checklist


iamfrankiemoran (approved):

ChrisGarratt122 (approved):

src/app/app.component.html
```diff
@@ -34,8 +34,13 @@
         @if (minutesRemaining$ | async; as minutes) {
           @if (thresholdInMinutes >= minutes && minutes > 0) {
             <div>
+              <opal-lib-custom-deferred-live-region-announcement
+                message="Warning: Your session will expire soon."
+                role="alert"
+              ></opal-lib-custom-deferred-live-region-announcement>
+
               @let minutesText = minutes === 1 ? 'minute' : 'minutes';
-              <div opal-lib-moj-alert ariaLabel="Your session will expire" type="warning">
```
Arnabsubedi233: Just curious, is it not working with the UI common lib component implementation, where the               'opal-lib-custom-deferred-live-region-announcement' is passed into the alert component. and you just have to pass the ariaLabel through?

Arnabsubedi233 (commented):
Small comment

src/app/app.component.html
sarahtw-cgi: It does work with the alert component and ariaLabel, but because the countdown updates every minute, the alert can get reanounced. Using the deferred live region lets us announce the warning once when it appears, while the visible countdown can continue updating without triggering unexpected behaviour.

Arnabsubedi233: Ah okay yeah that makes sense

Arnabsubedi233 (approved):
LGTM

ChrisGarratt122 (approved):
LGTM

src/app/app.component.html
Arnabsubedi233: should the message not say "Warning: Your session has expired. Please log back in."

src/app/app.component.html
```diff
@@ -34,8 +34,13 @@
         @if (minutesRemaining$ | async; as minutes) {
           @if (thresholdInMinutes >= minutes && minutes > 0) {
             <div>
+              <opal-lib-custom-deferred-live-region-announcement
+                message="Warning: Your session will expire soon."
```
Arnabsubedi233: It looks like the rest of the message is not being announced , 'Please save your work and log out, then log in back to continue' . I think we need this to also be announced.

Arnabsubedi233 (changes requested):
Just small comment from QA

src/app/app.component.html
sarahtw-cgi: That message is fine as the expired warning is static. It doesn't have the same issue of the visible text updating and competing with the ariaLabel. The deferred live region is only needed for the countdown warning to avoid those conflicting/repeated announcements.

src/app/app.component.html
sarahtw-cgi: I have now updated these to better mirror the visible text

Arnabsubedi233 (approved):
LGTM

Arnabsubedi233 (approved):
LGTM

